PR #4, opened 2022-05-09
cloud-ui

neurosnap:
https://app.shortcut.com/aptible/story/3553/migrate-cloud-ui-to-reploy-ui

Migrating code from [cloud-ui](https://github.com/aptible/cloud-ui)


.github/workflows/tests.yml
```diff
@@ -12,8 +12,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: v14.17.0
-          username: ${{secrets.APTIBLE_ROBOT_USERNAME}}
-          password: ${{secrets.APTIBLE_ROBOT_PASSWORD}}
+          node-version: v16.15.0
```
madhuravius, line 15: should be able to do 16.x so you don't have to increment minors if possible.  in some places this is pinned but in other places (Dockerfile, etc it's 16, so you should either use a `.x` or a pinned version consistently imo)

i'm not 100% sure about `.node-version`, but should behave similarly

yarn.lock
```diff
@@ -260,6 +260,16 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
+"@headlessui/react@^1.6.1":
```
madhuravius, line 263: can't comment on line but there's a yarn lockfile v2 stable now, right?

https://dev.to/arcanis/introducing-yarn-2-4eh1

src/ui/shared/menu-button.tsx
```diff
@@ -0,0 +1,55 @@
...
+} from '@reach/menu-button';
+
+import './menu.css';
+import { tokens } from './tokens';
+
+import cn from 'classnames';
+import { Size } from './button';
+
+export const Menu = (props: MenuProps) => (
+  <ReachMenu className={cn()} {...props} />
```
madhuravius, line 25: intentional empty cn? and below of empty strings?

src/ui/shared/detail-page-section.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <div>
+      <div className="flex flex-row items-center">
+        <div className="flex-1">
+          <div className={cn(tokens.type.h3)}>{title}</div>
+          <p className="mt-2 text-sm text-gray-700">{description}</p>
+        </div>
+        <div>
+          {actions &&
+            actions.map((a, i) => <Fragment key={`action-${i}`}>{a}</Fragment>)}
```
madhuravius, line 36: nit/pet peeve. i think chars saved here just results in awkwardness in reading, i'd prefer if:

```jsx
actions.map((action, idx) => <Fragment key={`action-${idx}`}>{action}</Fragment>)}
```

also not sure if you want to just use optional chaining here. i forget if it's applicable in render cleanly, (but i've seen people accidentally put truthys and get weird text, but your type from `SectionProps` should save you from that i think)

src/ui/pages/settings/team-list-view.tsx
```diff
@@ -0,0 +1,66 @@
+import type { User } from '@app/types';
+
+import {
+  TableHead,
+  Td,
+  Button,
+  tokens,
+  ResourceListView,
+  Input,
+} from '../../shared';
```
madhuravius, line 10: we may want to use aliases so we can avoid relative paths, i would prefer if possible to just have relative to root

src/ui/layouts/detail-page.tsx
```diff
@@ -0,0 +1,215 @@
...
+          <main className={cn(tokens.layout['main width'], 'py-6')}>
+            {children}
+          </main>
+        </div>
+      </div>
+    </>
+  );
+}
+
+/* function Overview() {
```
madhuravius, line 31: commented section, i'm sure it's known but just reminder

madhuravius:
left assorted notes, feel free to ignore, but will serve as placeholder in some discussion we have later in the week

PR #7, opened 2022-12-05
initial icons

grrowe:
- icon file exporting all icons
- swapped `icn` out for icons
- added magnifying glass to search bar

### DIFF
--- a/src/ui/shared/app/app-list.tsx
+++ b/src/ui/shared/app/app-list.tsx
@@ -122,18 +122,8 @@ export function AppList() {
         title="Apps"
         description={description}
         filterBar={
-          <div
-            className="flex flex-1 pt-4 gap-3"
-            style={{ position: "relative", margin: "5px" }}
-          >
-            <MagnifyingGlass
-              style={{
-                position: "absolute",
-                display: "inline-block",
-                top: "22",
-                left: "6",
-              }}
-            />
+          <div className="flex flex-1 pt-4 gap-3 relative m-1">
+            <MagnifyingGlass className="absolute inline-block top-6 left-1.5" />
             <Input
               placeholder="Search Apps ..."
               type="text"
--- a/src/ui/shared/app/app-list.tsx
+++ b/src/ui/shared/app/app-list.tsx
@@ -13,6 +13,7 @@ import type { AppState } from "@app/types";
 import { selectServicesByIds } from "@app/deploy";
 import { calcMetrics } from "@app/deploy";
 import { appDetailUrl } from "@app/routes";
+import { MagnifyingGlass } from "@app/ui/shared/icons";
 
 import { LoadResources } from "../../shared";
 import { TableHead, Td, tokens, ResourceListView, Input } from "../../shared";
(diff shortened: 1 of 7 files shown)


src/ui/shared/icons.tsx
```diff
@@ -0,0 +1,1413 @@
+export const SmallArrowUp = ({
+  color = "#111920",
+  width = "24",
+  height = "24",
+  style = {},
+}: {
```
neurosnap, line 6: Let's abstract this into an interface
```ts
interface IconProps {
  color?: string;
  // ...
}
```

src/ui/shared/endpoints-overview.tsx
```diff
@@ -81,7 +83,12 @@ export function EndpointsOverview({
             <EmptyResultView
               title="No endpoints yet"
               description="Expose this application to the public internet by adding an endpoint"
-              action={<Button variant="primary">Add Endpoint</Button>}
+              action={
+                <Button variant="primary">
+                  <PlusCircle color="#ffffff" />{" "}
```
neurosnap, line 88: I think it could be nicer to pre-determine the colors and then ask the end-dev to provide the name.  So instead of hex we would supply "white"

neurosnap (changes requested):

src/ui/shared/endpoints-overview.tsx
grrowe: changed to "white"

src/ui/shared/icons.tsx
grrowe: interface IconProps {
  color?: string;
  width?: string;
  height?: string;
  style?: object;
}

neurosnap (changes requested):
Very close!

src/ui/shared/app/app-list.tsx
```diff
@@ -121,12 +122,24 @@ export function AppList() {
         title="Apps"
         description={description}
         filterBar={
-          <div className="flex flex-1 pt-4 gap-3">
+          <div
+            className="flex flex-1 pt-4 gap-3"
+            style={{ position: "relative", margin: "5px" }}
```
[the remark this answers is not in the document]
grrowe: Sorry, keep forgetting about this. Updated.

src/ui/shared/app/app-list.tsx
```diff
@@ -121,12 +122,24 @@ export function AppList() {
...
+          >
+            <MagnifyingGlass
+              style={{
```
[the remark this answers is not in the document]
grrowe: There is a top, values are a bit different they use rem instead of px.

src/ui/shared/icons.tsx
```diff
@@ -0,0 +1,1337 @@
...
+}
+
+export const SmallArrowUp = ({
+  color = "#111920",
+  width,
+  height,
+  style = {},
+}: IconProps) => {
+  return (
+    <svg
```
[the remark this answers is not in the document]
grrowe: Created 2 different parent component, one for the stroke color icons and one for the fill color icons.

src/ui/shared/icons.tsx
```diff
@@ -0,0 +1,739 @@
...
+      height="24px"
+      fill={color}
+      className={`${className} w-6 h-6`}
+    >
+      {path}
+    </svg>
+  );
+};
+
+// small arrow up
```
neurosnap, line 58: I'm curious, what's the purpose of these comments?

neurosnap (approved):
I'd probably delete those comments but it's not a blocker

src/ui/shared/icons.tsx
grrowe: just to keep track, no real purpose besides readability.

PR #11, opened 2023-01-06
refactor icon props components for linter

grrowe:
- add props interface
- update `any` -> `Props`
    - optional props like `color`
- described component further with `React.FC<Props>`

### DIFF
--- a/src/ui/shared/icons.tsx
+++ b/src/ui/shared/icons.tsx
@@ -12,8 +12,6 @@ interface IconProps {
 const IconStrokeBase = ({
   path,
   color = "#111920",
-  width,
-  height,
   style = {},
   className = "",
 }: IconProps) => {
@@ -21,10 +19,7 @@ const IconStrokeBase = ({
     <svg
       style={style}
       xmlns="http://www.w3.org/2000/svg"
-      width={width}
-      height={height}
       fill="none"
-      viewBox="0 0 24 24"
       stroke={color}
       className={`${className} w-6 h-6`}
     >
@@ -35,8 +30,6 @@ const IconStrokeBase = ({
 const IconFillBase = ({
   path,
   color = "#111920",
-  width,
-  height,
   style = {},
   className = "",
 }: IconProps) => {
@@ -45,8 +38,6 @@ const IconFillBase = ({
     <svg
       style={style}
       xmlns="http://www.w3.org/2000/svg"
-      width={width}
-      height={height}
       fill={color}
       className={`${className} w-6 h-6`}
     >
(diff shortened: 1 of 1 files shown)


src/ui/shared/icons.tsx
```diff
@@ -55,8 +55,16 @@ const IconFillBase = ({
...
+  color?: string;
+  width?: string;
+  height?: string;
+  style?: object;
+  className?: string;
+}
+
 // small arrow up
-export const SmallArrowUp = (props: any) => {
+export const SmallArrowUp: React.FC<Props> = (props: Props) => {
```
neurosnap, line 67: > In most cases, it makes very little difference which syntax is used, but the React.FC syntax is slightly more verbose without providing clear advantage, so precedence was given to the "normal function" syntax.

https://stackoverflow.com/a/59991281

It seems like we don't need `React.FC<Props>` so I would probably not add it.

src/ui/shared/icons.tsx
```diff
@@ -55,8 +55,16 @@ const IconFillBase = ({
   );
 };
 
+export interface Props extends React.HTMLProps<HTMLDivElement> {
```
neurosnap, line 58: I think this should be `React.SVGProps<SVGSVGElement>`: https://stackoverflow.com/a/60083905

src/ui/shared/icons.tsx
```diff
@@ -120,7 +128,7 @@ export const ArrowsUpDown = (props: any) => {
 };
 
 // <!-- chevron-up -->
-export const ChevronUp = (props: any) => {
+export const ChevronUp: React.FC<Props> = (props: Props) => {
```
neurosnap, line 131: There are cases where you aren't using `props`.  Please confirm all icons are using them.

neurosnap (changes requested):
`IconFillBase` has `width` and `height` but it is never used in the component.

src/ui/shared/icons.tsx
grrowe: confirmed, all icons are using props

src/ui/shared/icons.tsx
grrowe: You're right, swapped it over.

src/ui/shared/icons.tsx
grrowe: Removed

src/ui/shared/icons.tsx
```diff
@@ -45,8 +45,8 @@ const IconFillBase = ({
     <svg
       style={style}
       xmlns="http://www.w3.org/2000/svg"
-      width="24px"
-      height="24px"
+      width={width}
```
neurosnap, line 48: Do we ever change the width/height?  If not let's remove the option.

grrowe: We do not currently, but I am not sure if we will in the future or not. I will remove it now and just add it back in if we need it later.

neurosnap (approved):

PR #13, opened 2023-02-03
feat(ftux): entire flow mostly functional

neurosnap:
A user can now traverse the entire flow of our new first time user experience.

`/create` -> new endpoint created for ftux flow

A user can:
- signup / login / be logged in
- add ssh key
- create an environment and an app
- deploy app
- create and provision (n) databases
- configure environment variables
- monitor all operations

What's left:
- Check for scan operation
- Add service commands
- Add endpoints
- Show logs after op completed
- Deployment failed screen
- Review deployment screen

https://user-images.githubusercontent.com/1940365/217033912-dbc4bf54-1788-4d6b-bce8-f2e37a0d0868.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "@tailwindcss/typography": "^0.5.9",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
+    "@types/debug": "^4.1.7",
     "@types/node": "^18.11.18",
     "@types/react": "^18.0.25",
     "@types/react-dom": "^18.0.9",
@@ -40,10 +41,12 @@
     "autoprefixer": "^10.4.13",
     "classnames": "^2.3.2",
     "date-fns": "^2.29.3",
+    "debug": "^4.3.4",
     "history": "^5.3.0",
     "husky": "^8.0.3",
     "jsdom": "^21.1.0",
     "lint-staged": "^13.1.0",
+    "nanoid": "^4.0.0",
     "postcss": "^8.4.21",
     "qrcode.react": "^3.1.0",
     "query-string": "^8.1.0",
@@ -62,5 +65,6 @@
     "vite": "^4.0.4",
     "vite-tsconfig-paths": "^4.0.5",
     "vitest": "^0.28.2"
-  }
+  },
+  "devDependencies": {}
 }
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -10,6 +10,7 @@ import type { AppState } from "@app/types";
 import { resetReducer } from "@app/reset-store";
 import { TOKEN_NAME, ELEVATED_TOKEN_NAME } from "@app/token";
 import { THEME_NAME } from "@app/theme";
+import { REDIRECT_NAME } from "@app/redirect-path";
 
 import { sagas, reducers } from "./packages";
 
@@ -25,7 +26,7 @@ interface AppStore<State> {
 const persistConfig = {
   key: "root",
   storage,
-  whitelist: [TOKEN_NAME, ELEVATED_TOKEN_NAME, THEME_NAME],
+  whitelist: [TOKEN_NAME, ELEVATED_TOKEN_NAME, THEME_NAME, REDIRECT_NAME],
 };
 
 export function setupStore({ initState }: Props): AppStore<AppState> {
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -4,8 +4,9 @@ import type { Middleware, Store } from "@reduxjs/toolkit";
 import { PersistPartial } from "redux-persist/es/persistReducer";
 import { persistStore, persistReducer } from "redux-persist";
 import storage from "redux-persist/lib/storage"; // defaults to localStorage for web
-import { prepareStore } from "saga-query";
+import { BATCH, prepareStore } from "saga-query";
 
+import { createLog } from "@app/debug";
 import type { AppState } from "@app/types";
 import { resetReducer } from "@app/reset-store";
 import { TOKEN_NAME, ELEVATED_TOKEN_NAME } from "@app/token";
@@ -29,22 +30,24 @@ const persistConfig = {
   whitelist: [TOKEN_NAME, ELEVATED_TOKEN_NAME, THEME_NAME, REDIRECT_NAME],
 };
 
+const log = createLog("redux");
+
 export function setupStore({ initState }: Props): AppStore<AppState> {
   const middleware: Middleware[] = [];
 
   if (import.meta.env.VITE_DEBUG === "true") {
-    /* const logger = (store: any) => (next: any) => (action: any) => {
+    const logger = (store: any) => (next: any) => (action: any) => {
       if (action.type === BATCH) {
-        console.log('== BATCH ==');
+        log("== BATCH ==");
         action.payload.forEach(console.log);
-        console.log('== END BATCH ==');
+        log("== END BATCH ==");
       } else {
-        console.log('ACTION', action);
+        log("ACTION", action);
       }
       next(action);
-      console.log('NEXT STATE', store.getState());
+      log("NEXT STATE", store.getState());
     };
-    middleware.push(logger); */
+    middleware.push(logger);
   }
 
   const prepared = prepareStore({
--- a/src/auth/signup.ts
+++ b/src/auth/signup.ts
@@ -6,6 +6,7 @@ import {
   setLoaderError,
 } from "saga-query";
 
+import { createLog } from "@app/debug";
 import { AuthApiCtx, ThunkCtx, thunks } from "@app/api";
 import { CreateUserForm, CreateUserCtx, createUser } from "@app/users";
 
@@ -13,7 +14,8 @@ import { TokenCtx, createToken, elevateToken, ElevateTokenCtx } from "./token";
 import { AUTH_LOADER_ID } from "./loader";
 import { createOrganization, OrgCtx } from "@app/organizations";
 import { ApiGen } from "@app/types";
-import { elevate } from "./elevate";
+
+const log = createLog("signup");
 
 function* setAuthError(ctx: AuthApiCtx) {
   if (ctx.json.ok) {
@@ -33,7 +35,9 @@ export const signup = thunks.create<CreateUserForm>(
       createUser.run,
       createUser(ctx.payload),
     );
-    console.log(userCtx);
+
+    log(userCtx);
+
     if (!userCtx.json.ok) {
       yield call(setAuthError, userCtx);
       return;
@@ -48,7 +52,9 @@ export const signup = thunks.create<CreateUserForm>(
         makeCurrent: true,
       }),
     );
-    console.log(tokenCtx);
+
+    log(tokenCtx);
+
     if (!tokenCtx.json.ok) {
       yield call(setAuthError, tokenCtx);
       return;
@@ -58,7 +64,9 @@ export const signup = thunks.create<CreateUserForm>(
       createOrganization.run,
       createOrganization({ name: email }),
     );
-    console.log(orgCtx);
+
+    log(orgCtx);
+
     if (!orgCtx.json.ok) {
       yield call(setAuthError, orgCtx);
       return;
@@ -68,7 +76,8 @@ export const signup = thunks.create<CreateUserForm>(
       elevateToken.run,
       elevateToken({ username: email, password, otpToken: "" }),
     );
-    console.log(elevateCtx);
+
+    log(elevateCtx);
 
     yield put(
       setLoaderSuccess({
--- a/src/auth/signup.ts
+++ b/src/auth/signup.ts
@@ -9,8 +9,11 @@ import {
 import { AuthApiCtx, ThunkCtx, thunks } from "@app/api";
 import { CreateUserForm, CreateUserCtx, createUser } from "@app/users";
 
-import { TokenCtx, createToken } from "./token";
+import { TokenCtx, createToken, elevateToken, ElevateTokenCtx } from "./token";
 import { AUTH_LOADER_ID } from "./loader";
+import { createOrganization, OrgCtx } from "@app/organizations";
+import { ApiGen } from "@app/types";
+import { elevate } from "./elevate";
 
 function* setAuthError(ctx: AuthApiCtx) {
   if (ctx.json.ok) {
@@ -22,7 +25,7 @@ function* setAuthError(ctx: AuthApiCtx) {
 
 export const signup = thunks.create<CreateUserForm>(
   "signup",
-  function* onSignup(ctx: ThunkCtx<CreateUserForm>, next) {
+  function* onSignup(ctx: ThunkCtx<CreateUserForm>, next): ApiGen {
     const { email, password } = ctx.payload;
     yield put(setLoaderStart({ id: AUTH_LOADER_ID }));
 
@@ -51,7 +54,31 @@ export const signup = thunks.create<CreateUserForm>(
       return;
     }
 
-    yield put(setLoaderSuccess({ id: AUTH_LOADER_ID }));
+    const orgCtx: OrgCtx = yield call(
+      createOrganization.run,
+      createOrganization({ name: email }),
+    );
+    console.log(orgCtx);
+    if (!orgCtx.json.ok) {
+      yield call(setAuthError, orgCtx);
+      return;
+    }
+
+    const elevateCtx: ElevateTokenCtx = yield call(
+      elevateToken.run,
+      elevateToken({ username: email, password, otpToken: "" }),
+    );
+    console.log(elevateCtx);
+
+    yield put(
+      setLoaderSuccess({
+        id: AUTH_LOADER_ID,
+        meta: {
+          id: userCtx.json.data.id,
+          verified: userCtx.json.data.verified,
+        },
+      }),
+    );
     yield next();
   },
 );
--- a/src/deploy/app/index.ts
+++ b/src/deploy/app/index.ts
@@ -202,7 +202,7 @@ export const createDeployApp = api.post<CreateAppProps>(
 
 interface CreateAppOpProps {
   appId: string;
-  env: { [key: string]: string | boolean | number };
+  env: { [key: string]: string };
 }
 export type CreateAppOpCtx = DeployApiCtx<any, CreateAppOpProps>;
 export const createAppOperation = api.post<CreateAppOpProps>(
--- a/src/deploy/app/index.ts
+++ b/src/deploy/app/index.ts
@@ -1,11 +1,23 @@
-import { api, cacheTimer, combinePages, PaginateProps, thunks } from "@app/api";
+import {
+  api,
+  cacheTimer,
+  combinePages,
+  DeployApiCtx,
+  PaginateProps,
+  thunks,
+} from "@app/api";
 import { defaultEntity, extractIdFromLink } from "@app/hal";
 import {
   createReducerMap,
   createTable,
   mustSelectEntity,
 } from "@app/slice-helpers";
-import type { DeployApp, AppState } from "@app/types";
+import type {
+  DeployApp,
+  AppState,
+  LinkResponse,
+  ProvisionableStatus,
+} from "@app/types";
 import { createSelector } from "@reduxjs/toolkit";
 
 import { selectEnvironments, findEnvById } from "../environment";
@@ -15,7 +27,28 @@ import { selectDeploy } from "../slice";
 
 export * from "./utils";
 
-export const deserializeDeployApp = (payload: any): DeployApp => {
+export interface DeployAppResponse {
+  id: string;
+  handle: string;
+  git_repo: string;
+  created_at: string;
+  updated_at: string;
+  deployment_method: string;
+  status: ProvisionableStatus;
+  _links: {
+    account: LinkResponse;
+    current_configuration: LinkResponse;
+  };
+  _embedded: {
+    // TODO: fill in
+    services: { id: string }[];
+    current_image: any;
+    last_deploy_operation: any;
+    last_operation: any;
+  };
+}
+
+export const deserializeDeployApp = (payload: DeployAppResponse): DeployApp => {
   const serviceIds: string[] = payload._embedded.services.map((s: any) => s.id);
   const links = payload._links;
   const embedded = payload._embedded;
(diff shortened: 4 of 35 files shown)


src/app/store.ts
```diff
@@ -25,7 +26,7 @@ interface AppStore<State> {
 const persistConfig = {
   key: "root",
   storage,
-  whitelist: [TOKEN_NAME, ELEVATED_TOKEN_NAME, THEME_NAME],
+  whitelist: [TOKEN_NAME, ELEVATED_TOKEN_NAME, THEME_NAME, REDIRECT_NAME],
 };
 
 export function setupStore({ initState }: Props): AppStore<AppState> {
```
madhuravius, line 32: if debug i don't see the harm in leaving this in (lines 34 - 48)

if unusable, i'd drop

src/auth/signup.ts
```diff
@@ -9,8 +9,11 @@ import {
 import { AuthApiCtx, ThunkCtx, thunks } from "@app/api";
 import { CreateUserForm, CreateUserCtx, createUser } from "@app/users";
 
-import { TokenCtx, createToken } from "./token";
+import { TokenCtx, createToken, elevateToken, ElevateTokenCtx } from "./token";
```
madhuravius, line 12: nit/minor note, we should have a linter that prevents absolute imports and relative imports, i would force absolute everywhere (even in the same dir) if possible

src/auth/signup.ts
```diff
@@ -51,7 +54,31 @@ export const signup = thunks.create<CreateUserForm>(
       return;
     }
 
-    yield put(setLoaderSuccess({ id: AUTH_LOADER_ID }));
+    const orgCtx: OrgCtx = yield call(
+      createOrganization.run,
+      createOrganization({ name: email }),
+    );
+    console.log(orgCtx);
```
madhuravius, line 61: might wanna drop console logs

src/deploy/app/index.ts
```diff
@@ -15,7 +27,28 @@ import { selectDeploy } from "../slice";
 
 export * from "./utils";
 
-export const deserializeDeployApp = (payload: any): DeployApp => {
+export interface DeployAppResponse {
```
madhuravius, line 30: are these all required? i think i'd verify these with a `?:` for the optional fields (i think that's the notation?)

src/deploy/app/index.ts
```diff
@@ -15,7 +27,28 @@ import { selectDeploy } from "../slice";
...
+  created_at: string;
+  updated_at: string;
+  deployment_method: string;
+  status: ProvisionableStatus;
+  _links: {
+    account: LinkResponse;
+    current_configuration: LinkResponse;
+  };
+  _embedded: {
+    // TODO: fill in
```
madhuravius, line 43: just reminder on TODO unless intended for later, might be safe to drop

src/deploy/app/index.ts
```diff
@@ -15,7 +27,28 @@ import { selectDeploy } from "../slice";
...
+    current_image: any;
+    last_deploy_operation: any;
+    last_operation: any;
+  };
+}
+
+export const deserializeDeployApp = (payload: DeployAppResponse): DeployApp => {
   const serviceIds: string[] = payload._embedded.services.map((s: any) => s.id);
   const links = payload._links;
   const embedded = payload._embedded;
```
madhuravius, line 54: line 89 at the bottom of this is a spread operator with possible deep comparisons needed (you may need to do copies of service ids and/or other nested objects) to be safe

madhuravius: line 139-153 should probably be ternary with a const as they are not really reassigned if the conditional parent is set

madhuravius: near the bottom, should probably do falsy checks, not !== "" in JS (we probably should have a linter that yells about this)

src/ui/pages/elevate.tsx
```diff
@@ -113,6 +118,7 @@ export const ElevatePage = () => {
                     autoComplete="off"
                     id="input-2fa"
                     className="flex-1 outline-0 py-1 bg-transparent"
+                    autoFocus
```
madhuravius, line 121: might want to be cautious of this field unless desired. is there a reason to set this if there are other inputs on the page that take precedence? (is it for convenience for users that use password managers?)

src/ui/pages/signup.tsx
```diff
@@ -68,29 +72,33 @@ const SignupPageForm = () => {
...
-    password !== "" && passwordErrors.length > 0
-      ? `Password ${passwordErrors.join(", ")}`
-      : "";
-  const disableSave =
-    name === "" ||
-    currentEmail === "" ||
-    password === "" ||
-    !!emailErrorMessage ||
-    passwordErrors.length > 0;
+  const disableSave = name === "" || currentEmail === "" || password === "";
```
madhuravius, line 75: is there a need to do this? if falsy should be sufficient if we have types enforced, right?

src/ui/pages/signup.tsx
```diff
@@ -105,6 +113,13 @@ const SignupPageForm = () => {
     if (invitationRequest.invitationId) {
       navigate(acceptInvitationWithCodeUrl(invitationRequest));
     } else {
+      // if the api returns with a user.verified = true, skip email request page
+      // this can happen in development when ENV['DISABLE_EMAIL_VERIFICATION']=1
+      if (loader.meta.verified) {
```
madhuravius, line 118: is `meta.verified` guaranteed? may want to optionally chain

src/ui/pages/signup.tsx
```diff
@@ -161,24 +165,11 @@ const SignupPageForm = () => {
...
-                />
-              </FormGroup>
-
               <FormGroup
-                label="Your work email"
+                label="Your email"
                 htmlFor="email"
-                feedbackVariant={emailErrorMessage ? "danger" : "info"}
-                feedbackMessage={emailErrorMessage}
+                feedbackVariant={emailError ? "danger" : "info"}
```
madhuravius, line 171: to be consistent, may want to use === "" or stick with falsy checks in general for consistency. a linter should probably catch this type of thing imo

package.json
```diff
@@ -62,5 +65,6 @@
     "vite": "^4.0.4",
     "vite-tsconfig-paths": "^4.0.5",
     "vitest": "^0.28.2"
-  }
+  },
+  "devDependencies": {}
```
madhuravius, line 69: nit - but, intentional? usually all the @types go here and/or build/test tools (not that it matters too much as we're transpiling)

src/deploy/app/index.ts
```diff
@@ -144,6 +177,44 @@ export const fetchAppOperations = api.get<{ id: string }>(
...
+      body: JSON.stringify(body),
+    });
+
+    yield next();
+  },
+);
+
+interface CreateAppOpProps {
+  appId: string;
+  env: { [key: string]: string | boolean | number };
```
madhuravius, line 205: is this really true? usually an env is a key value map of strings only, right? maybe it's set differently on aptible backend somehow? (and ruby maps on backend are handling this differently?)

src/deploy/app/index.ts
```diff
@@ -144,6 +177,44 @@ export const fetchAppOperations = api.get<{ id: string }>(
...
+}
+export type CreateAppOpCtx = DeployApiCtx<any, CreateAppOpProps>;
+export const createAppOperation = api.post<CreateAppOpProps>(
+  "/apps/:appId/operations",
+  function* (ctx: CreateAppOpCtx, next) {
+    const { env } = ctx.payload;
+    const body = { env, type: "configure" };
```
madhuravius, line 212: nit, but i prefer doing this kind of thing:

1. avoiding an intermediate variable as there's no need and simply: 

```js
ctx.request = ctx.req({ body: JSON.stringify({ env, type: "configure" }) });
```

OR (preferred)

```js
const body = JSON.stringify({ env, type: "configure" });
ctx.request = ctx.req({ body });
```

madhuravius: i prefer 2 when destructuring assignment is commonplace in the code (for imports like you're using) as it seems more stylistically consistent

madhuravius:
i left a lot of comments we can chat through (feel free to disagree with any of them), but I would prefer stylistic consistency enforced by a linter.

🌶️  take - i see a lot of serialization/deserialization in what seems to mostly amount to de-hal'ing and snake case <--> pascal case. if there's minimal changes, why not just stick with snake_case (as we're likely to enforce that on the backend).  is there a compelling reason to deviate from that for some reason?

src/deploy/database-images/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    visible: payload.visible,
+    createdAt: payload.created_at,
+    updatedAt: payload.updated_at,
+  };
+};
+
+export const defaultDeployDatabaseImage = (
+  db: Partial<DeployDatabaseImage> = {},
+): DeployDatabaseImage => {
+  const now = new Date().toISOString();
```
madhuravius, line 43: do we rely on utc server side or will this be wrong/discarded?

src/ui/shared/add-ssh-key.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
+import { useLoader, useLoaderSuccess } from "saga-query/react";
+import { BannerMessages } from "./banner-messages";
+import { Button } from "./button";
+import { FormGroup } from "./form-group";
+import { Input } from "./input";
+import { tokens } from "./tokens";
+
+export const AddSSHKeyForm = ({
```
madhuravius, line 12: for some reason i cannot seem to find this file's usages anywhere, was this intended to be used? or accidentally committed?

is there ever a time we will not care about onSuccess? if so why not just make it optional and then do a conditional and check instead of a generic callback that serves no purpose? (feels odd)

src/ui/shared/add-ssh-key.tsx
```diff
@@ -0,0 +1,79 @@
...
+  useLoaderSuccess(loader, () => {
+    onSuccess();
+  });
+
+  return (
+    <div>
+      <h2 className={tokens.type.h3}>Public SSH Key</h2>
+      <p className="text-gray-600 mb-2">
+        Copy the contents of the file <code>$HOME/.ssh/id_rsa.pub</code> then
+        paste here. Need Help? View Docs
```
madhuravius, line 37: should probably be a link/href? (`View Docs` that is)

src/ui/layouts/auth-required.tsx
```diff
@@ -1,18 +1,29 @@
...
+  const authed = loader.isLoading || isAuthenticated;
 
-  if (!(loader.isLoading || isAuthenticated)) {
+  useEffect(() => {
+    if (!authed) {
+      dispatch(setRedirectPath(location.pathname));
+    }
+  }, [authed]);
+
+  if (!authed) {
```
madhuravius, line 26: i think i'm reading something wrong here.

if not authed on mount, it will redirect
after mount, if not authed, they will see this navigate to login page (or in initial load).

is there any reason to leave this in? won't they just see it flash and disappear or be redirected? (would feel glitchy or weird in general)

src/ui/pages/create-project-git.tsx
```diff
@@ -34,17 +95,24 @@ const FormNav = ({
...
   return (
     <div>
-      <Link aria-disabled={!prev} to={prev} className="pr-2">
-        Prev
-      </Link>
-      <Link aria-disabled={!next} to={next}>
-        Next
-      </Link>
+      {prev ? (
+        <Link aria-disabled={!prev} to={prev} className="pr-2">
```
madhuravius, line 99: consistency type issue as in rest of pr, equality check with `===""` or falsy types like here, important to remain consistent imo

src/ui/pages/create-project-git.tsx
```diff
@@ -34,17 +95,24 @@ const FormNav = ({
 }) => {
   return (
```
madhuravius, line 96: nit - if prev/next are both falsy, might want to return a null in its entirety as this `div` is unstyled it seems and doesn't need to be rendered

similarly, i'm not sure why:

```js
{!!prev && (<Link>...</Link)}
{!!next && (<Link>...</Link)}
```

is not sufficient here as it's more readable than needless ternaries, right?

src/ui/pages/create-project-git.tsx
```diff
@@ -54,27 +122,130 @@ export const CreateProjectAddKeyPage = () => {
...
-        <h2 className={tokens.type.h3}>Public SSH Key</h2>
-        <p className="text-gray-600 mb-2">
-          Copy the contents of the file <code>$HOME/.ssh/id_rsa.pub</code> then
-          paste here. Need Help? View Docs
-        </p>
-        <textarea className={tokens.type.textarea} />
-        <Button className="w-full mt-2" disabled>
-          Save Key
-        </Button>
+        <AddSSHKeyForm onSuccess={onSuccess} />
```
madhuravius, line 128: ah it's used here, for some reason this wasn't coming up in github search, sorry!

src/ui/pages/create-project-git.tsx
```diff
@@ -84,35 +255,179 @@ export const CreateProjectGitPushPage = () => {
...
+
+const validateDbs = (
+  items: TextVal[],
+  dbImages: DeployDatabaseImage[],
+): ValidatorError[] => {
+  const errors: ValidatorError[] = [];
+
+  const validate = (item: TextVal) => {
+    const imgs = dbImages.filter((img) => img.type === item.key);
+    if (imgs.length === 0) {
```
madhuravius, line 313: why not `imgs?.length`, is there a reason we will ever have a non nil/integer value for a length array type?

src/ui/pages/create-project-git.tsx
```diff
@@ -84,35 +255,179 @@ export const CreateProjectGitPushPage = () => {
...
+        item,
+        message: `[${item.key}] is not a valid database`,
+      });
+      return;
+    }
+
+    let found = false;
+    imgs.forEach((img) => {
+      if (img.version === item.value) {
+        found = true;
```
madhuravius, line 324: you probably want a break to avoid looping over the entire array once found

src/ui/pages/create-project-git.tsx
```diff
@@ -84,35 +255,179 @@ export const CreateProjectGitPushPage = () => {
...
+      }
+    });
+
+    if (!found) {
+      errors.push({
+        item,
+        message: `[${item.value}] is not a valid version for [${item.key}]`,
```
madhuravius, line 331: are we using something for the `[]` or does the bracket look different stylistically? feels a bit odd unless we style it somehow (or if it's intended)

src/ui/pages/create-project-git.tsx
```diff
@@ -84,35 +255,179 @@ export const CreateProjectGitPushPage = () => {
...
+
+const validateEnvs = (items: TextVal[]): ValidatorError[] => {
+  const errors: ValidatorError[] = [];
+
+  const validate = (item: TextVal) => {
+    // https://stackoverflow.com/a/2821201
+    if (!/[a-zA-Z_]+[a-zA-Z0-9_]*/.test(item.key)) {
+      errors.push({
+        item,
+        message: `${item.key} does not match regex: /[a-zA-Z_]+[a-zA-Z0-9_]*/`,
```
madhuravius, line 348: might want a parens explaining this to a normal person (Please only use strings, numbers, underscores or hyphens) as a regex error isn't the nicest thing in the world to see

madhuravius:
i left a fair number of comments/questions, mainly to start discussion/solidifying things:

* consistency (leveraging typing or consistent usage of falsy guards / empty checks, etc)
* not leaning on functional code choices that js usually does a good job
* some optional chaining bits
* questions on snake/pascal casing for the sake of something (not sure what that something is yet) and sometimes using the un-deserialized data references

overall lgtm, but left these questions here for us to discuss, resolve or declare a choice/opinion and remain consistent on it, let me know how you want to proceed/followup

src/ui/pages/create-project.tsx
```diff
@@ -1,26 +1,36 @@
...
+      <main className="flex-1">
+        <div className="py-6">
+          <div className="max-w-7xl mx-auto px-4 sm:px-6 md:px-8">
+            <div className="py-4">
+              <div className="flex justify-center container">
+                <div style={{ width: 500 }}>
+                  <div className="text-center">
+                    <h1 className={tokens.type.h1}>Create an Environment</h1>
+                    <p className="my-4 text-gray-600">
+                      Choose a deployment type for your app to get started.
```
madhuravius, line 16: maybe i'm not seeing it but it might be worthwhile to create a placeholder here so users know that there's another type coming

src/auth/signup.ts
madhuravius: or use that `log` helper i saw elsehwere in pr?

src/deploy/stack/index.ts
```diff
@@ -82,6 +82,17 @@ export const selectStacksAsOptions = createSelector(
     });
   },
 );
+export const selectStackPublicDefault = createSelector(
+  selectStacksAsList,
+  (stacks) => {
+    if (stacks.length === 0) {
```
madhuravius, line 88: probably doesn't need braces and can be put on the same line unless this is linter enforced?

similarly length should be a truthy/falsy and `0` is falsy, consistency in these choices would be good

src/auth/signup.ts
neurosnap: The general rule that has worked well in previous projects is this:
- Absolute import across packages
- Relative import inside packages

I can write a lint script to codify this rule.

src/ui/pages/elevate.tsx
neurosnap: After the user types in user/pass and then we display OTP field, we want that field to automatically be selected.

src/deploy/app/index.ts
neurosnap: Let's chat about this in a sync because I'm unclear what you are referencing.

src/ui/pages/signup.tsx
neurosnap: True, sometimes I float between being explicit and not.

src/ui/pages/signup.tsx
neurosnap: `loader.meta` is guaranteed to be an object so worst case it'll resolve to `undefined` which will fail the if-statement.

src/ui/pages/signup.tsx
neurosnap: `emailError` in this case is a `boolean`.

package.json
neurosnap: This is a weird situation with FE apps that get bundled.  Everything is a `devDependency` because we generate a bundle that includes everything, so there is no `dependencies`.  However, if you omit `--dev` then it gets installed as a dependency.  People inevitably forget the `--dev` flag and it causes all sorts of confusion when one slips through review.

I find it easier to just put everything inside `dependencies` because it is a meaningless distinction for FE apps.

src/deploy/app/index.ts
neurosnap: You're right, I'll change this.

src/deploy/app/index.ts
neurosnap: In my next PR I have refactored this.

src/ui/shared/add-ssh-key.tsx
neurosnap: Yep, Joe and I are going to do a pass on this.

src/ui/layouts/auth-required.tsx
neurosnap: We need this because every page gets wrapped by this logic.  Sometimes when we make a request the response comes back with "token expired" so we always want to check for authentication.  This makes it so we always check and redirect if necessary.

Does that make sense?

src/ui/pages/create-project-git.tsx
neurosnap: In my next PR I refactored this to match your request.

src/ui/pages/create-project-git.tsx
neurosnap: When you perform a `.filter` it will always return an array.

src/ui/pages/create-project-git.tsx
neurosnap: I'd have to convert it to a traditional `for-loop` and figured the ergonomics out-weight the perf improvement.  I did battle this thought tho when implementing.

src/ui/pages/create-project-git.tsx
neurosnap: I tend to wrap variables inside brackets so it's clear what the exact variable is.

src/ui/pages/create-project-git.tsx
neurosnap: These are going to be developers so I figured it would be fine this way.  Joe will have his take on it so I'll see what he thinks.

src/ui/pages/create-project-git.tsx
```diff
@@ -84,35 +255,179 @@ export const CreateProjectGitPushPage = () => {
...
+      });
+    }
+  };
+
+  items.forEach(validate);
```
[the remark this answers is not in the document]
neurosnap: Feels cleaner to me the way I wrote it.

src/deploy/stack/index.ts
neurosnap: For lengths I always choose explicit, not sure if it matters much, but it makes more sense to me.

package.json
madhuravius: sounds good, as long as that's our intention, wasn't sure  if it was, and yeah FE is weird for this stuff because it doesn't really matter as much

src/ui/shared/add-ssh-key.tsx
madhuravius: sounds good, if we always want a onSuccess, is the `:?` needed? isn't it a `:`?

src/ui/pages/elevate.tsx
madhuravius: TIL, assumed autofocus would immediately cause user focus regardless of position on page (assuming no other targets specified), using this can be weird behavior but i have no strong preference on this for now

src/ui/pages/create-project-git.tsx
madhuravius: got it, makes sense, just didn't see it elsewhere

src/ui/pages/create-project-git.tsx
madhuravius: probably brainfarted here, i think i was trying to say something else but doublebang instead of === 0, which is the same and mixed up thoughts

madhuravius (approved):

src/auth/signup.ts
madhuravius: if lintable, that'd be nice, i'm okay with this distinction

src/ui/pages/create-project.tsx
neurosnap: Users won't be seeing it in this state so I don't really see the point.  We'll be quickly adding more types soon.

PR #14, opened 2023-02-09
Code scanning

neurosnap:
This commit also:

- Displays results of code scan on the create git settings page
- Save all operations inside redux as its own separate entity (as opposed to using auto-caching)
    - And create selectors to pull data from that redux slice
- Created a discriminated union for the `createAppOperation` endpoint
- Adds `app:deploy` as an operation for `createAppOperation`
- Uses `app:deploy` inside `deployProject`

### DIFF
--- a/src/deploy/app/index.ts
+++ b/src/deploy/app/index.ts
@@ -18,11 +18,12 @@ import type {
   LinkResponse,
   ProvisionableStatus,
 } from "@app/types";
-import { createSelector } from "@reduxjs/toolkit";
+import { createAction, createSelector } from "@reduxjs/toolkit";
+import { poll } from "saga-query";
 
 import { selectEnvironments, findEnvById } from "../environment";
 import { deserializeImage } from "../image";
-import { deserializeOperation } from "../operation";
+import { deserializeDeployOperation } from "../operation";
 import { selectDeploy } from "../slice";
 
 export * from "./utils";
@@ -65,8 +66,12 @@ export const deserializeDeployApp = (payload: DeployAppResponse): DeployApp => {
     environmentId: extractIdFromLink(links.account),
     currentConfigurationId: extractIdFromLink(links.current_configuration),
     currentImage: deserializeImage(embedded.current_image),
-    lastDeployOperation: deserializeOperation(embedded.last_deploy_operation),
-    lastOperation: deserializeOperation(embedded.last_operation),
+    lastDeployOperation: embedded.last_deploy_operation
+      ? deserializeDeployOperation(embedded.last_deploy_operation)
+      : null,
+    lastOperation: embedded.last_operation
+      ? deserializeDeployOperation(embedded.last_operation)
+      : null,
   };
 };
 
@@ -167,13 +172,23 @@ export const fetchAllApps = thunks.create(
   combinePages(fetchApps),
 );
 
-export const fetchApp = api.get<{ id: string }>("/apps/:id", {
+interface AppIdProp {
+  id: string;
+}
+
+export const fetchApp = api.get<AppIdProp>("/apps/:id", {
   saga: cacheTimer(),
 });
 
-export const fetchAppOperations = api.get<{ id: string }>(
+export const fetchAppOperations = api.get<AppIdProp>(
   "/apps/:id/operations",
-  { saga: cacheTimer() },
+  api.cache(),
+);
+
+export const cancelAppOpsPoll = createAction("cancel-app-ops-poll");
+export const pollAppOperations = api.get<AppIdProp>(
+  ["/apps/:id/operations", "poll"],
+  { saga: poll(5 * 1000, `${cancelAppOpsPoll}`) },
   api.cache(),
 );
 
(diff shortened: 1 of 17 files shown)


src/deploy/app/index.ts
```diff
@@ -200,16 +215,47 @@ export const createDeployApp = api.post<CreateAppProps>(
   },
 );
 
-interface CreateAppOpProps {
+interface ScanAppOpProps {
```
madhuravius, line 218: this kind of repetition for `XOpProps` could be another thing this thing extends or is composed of, forget how to notate that but just a minor thing to look out for later

src/deploy/app/index.ts
```diff
@@ -200,16 +215,47 @@ export const createDeployApp = api.post<CreateAppProps>(
...
   function* (ctx: CreateAppOpCtx, next) {
-    const { env } = ctx.payload;
-    const body = { env, type: "configure" };
+    const getBody = () => {
+      const { type } = ctx.payload;
+      switch (type) {
+        case "deploy":
+        case "scan_code": {
+          const { gitRef } = ctx.payload;
+          return { type, git_ref: gitRef };
```
madhuravius, line 245: nit

can assign to value and do a break as well instead of returning in an anonymous function that inherits from parent context. 

similarly, might be worthwhile to move this out and set the values inside it for a case/switch so it's unit-testable instead of a generated func

src/deploy/operation/index.ts
```diff
@@ -27,18 +41,61 @@ export interface DeployOperationResponse {
   user_email: string;
   user_name: string;
   env: string;
+  _links: {
```
madhuravius, line 44: these are no joke, if there's a consistency it would be nice to pull all these into a common catch all (unless you really want it to be very strongly typed), where you have all the types and just pull from those keys 

not sure if it's this one (i think so?), but you could just specify an array since this is so repetitive of the keys you expect: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#key-remapping-via-as

src/hal/index.ts
```diff
@@ -17,7 +17,9 @@ import type { DeployApiCtx } from "@app/api";
   /([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})/,
 ); */
 
-export function extractIdFromLink(relation: { href: string } | null) {
+export function extractIdFromLink(
```
madhuravius, line 20: easy unit testable funcs here :D

src/hal/index.ts
```diff
@@ -28,6 +30,17 @@ export function extractIdFromLink(relation: { href: string } | null) {
...
 
+export function extractResourceNameFromLink(
+  resource: { href: string } | null | undefined,
+) {
+  if (!resource?.href) {
+    return "";
+  }
+
+  const res = resource.href.split("/");
+  return res[res.length - 2] || "";
```
madhuravius, line 41: might be safer to do `res?.[res.length - 2]` or something to avoid a white page/crash

something i realized is i didn't see a top level layout guard for page crashes (not sure if you added that yet, if not, just a reminder) whatever the equivalent of these are: https://reactjs.org/docs/error-boundaries.html

src/ui/pages/create-project-git.tsx
```diff
@@ -269,7 +287,16 @@ export const CreateProjectGitPushPage = () => {
           <h2 className={tokens.type.h3}>Push your code</h2>
           <PreCode>git push aptible main</PreCode>
         </div>
+
         <hr className="my-4" />
+
+        {deployOp ? (
+          <div>
+            We detected an app deployment, did you push to the{" "}
```
madhuravius, line 295: you have branch name in hand? or is this meant for something else? the `{" "}`

madhuravius: if intended to be a space, i believe &nbsp; is the way to do it, as awkward as that is (newline should do it by default)

src/ui/pages/create-project-git.tsx
```diff
@@ -443,6 +482,45 @@ export const CreateProjectGitSettingsPage = () => {
         next={createProjectGitStatusUrl(appId)}
       />
 
+      <Box>
+        {codeScan.isInitialLoading ? (
+          "Loading code scan results ...."
```
madhuravius, line 487: wrap this text in a paragraph? or some other dom element?

madhuravius: maybe like below? `<Loading text="Provisioning resources ..." />`

src/ui/pages/create-project-git.tsx
neurosnap: This is just what prettier does when you have a space in the sentence but need to wrap to nextline.

src/deploy/app/index.ts
neurosnap: > can assign to value and do a break as well instead of returning in an anonymous function that inherits from parent context.

I like the switch-case-to-return inside of a function style, what don't like you like about it?  The parent-child context is a non-issue because we are not using `this`, right?

> similarly, might be worthwhile to move this out and set the values inside it for a case/switch so it's unit-testable instead of a generated func

This seems low value.  It's already strongly typed so we aren't going to accidentally send an operation the wrong information based on its `type`.

If we did unit test this piece of code, it would provide us one main benefit:
- Ensure that the body is exactly what we expect it to, given a `type`

It doesn't confirm:
- That we are making the correct API request (url, method) with the proper payload (body)
- What we do with the API response
- That this function gets properly called when we dispatch `createOperation` with a valid payload

Further, writing unit tests for (pass `ctx.payload` and receive `body` object)  means everytime we change the payload for a given `type` we also have to change the test.  If we make a change to this code, it was most likely because we want to change the `body` -- which is intentional.  To me this is an example of a test that is:
- low impact (it's already strongly typed)
- busy work (everytime we make a change we also have to change the tests)

A higher impact test would be to test a user flow that involves `createOperation` with a test API server (e.g. msw) which would test that the entire stack is working properly (from user click to API requests made to what we do with the response, to what the user sees after success/failure).

src/ui/pages/create-project-git.tsx
madhuravius: TIL, mine always did the &nbsp; thing >10 months ago

src/deploy/app/index.ts
madhuravius: I'm okay with this, but I think this is a good conversation for us to have afterwards as it touches a lot of personal preferences that are different which is good to know:

* function's location and context (and how it is written), where a function's purpose is clear I/O without being manipulated by values outside its parameters (which leads to greater testability/readability, IMO)
* that functions do singular ideas of things (the confirmation list you added that wouldn't cover unit tests, which to me would be broken up and live elsewhere)

i agree with your assessment of utility/impact of the test addition here (it would be more of a write it/forget it for specific cases and add as you go type thing). either way this warrants a nice follow up convo

madhuravius (approved):

src/hal/index.ts
neurosnap: https://app.shortcut.com/aptible/story/9769/create-error-boundary-for-app

PR #15, opened 2023-02-10
Design app sidebar

neurosnap:
# BEFORE
Screen Shot 2023-02-10 at 9 54 54 AM

# AFTER
Screen Shot 2023-02-10 at 9 53 58 AM

### DIFF
--- a/index.css
+++ b/index.css
@@ -3,6 +3,7 @@
 @import "tailwindcss/utilities";
 
 html, body, #app {
+  font-size: 14px;
   width: 100%;
   height: 100%;
   overflow-x: hidden;
--- a/src/app/router.tsx
+++ b/src/app/router.tsx
@@ -38,6 +38,7 @@ import {
   CreateProjectGitSettingsPage,
   CreateProjectGitStatusPage,
   CreateProjectNamePage,
+  SettingsPageLayout,
 } from "@app/ui";
 import * as routes from "@app/routes";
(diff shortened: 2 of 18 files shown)


index.css
```diff
@@ -3,6 +3,7 @@
 @import "tailwindcss/utilities";
 
 html, body, #app {
+  font-size: 14px;
```
madhuravius, line 6: i'm accustomed to using em/rem and not hardcoding this (using scss vars)

maybe TODO/discussion for later, not actionable now

src/app/router.tsx
```diff
@@ -147,12 +148,17 @@ const appRoutes: RouteObject[] = [
...
-        path: routes.TEAM_PATH,
-        element: <TeamPage />,
+        element: <SettingsPageLayout />,
+        children: [
+          {
+            index: true,
+            element: <SettingsPage />,
+          },
+          {
+            path: routes.TEAM_PATH,
```
madhuravius, line 158: intended to use this layout?

neurosnap: Yep!

index.css
neurosnap: `rem` means: relative to the root `font-size`.  So we have to set an explicit font-size in pixels or else it defaults to 16px.

madhuravius (approved):

PR #17, opened 2023-02-14
feat(ftux): refactored status page

neurosnap:
Updated ftux git routes:

- before: `/create/git/:appId/status`
- after: `/apps/:appId/git/status`

Changed the logic for how we present operations on the status page.  Now we fetch all the resources within an account and then grab the latest operation that we are care about.

Did a design pass on the status page to match figma spec.

Screen Shot 2023-02-14 at 1 18 04 PM


src/hal/index.ts
```diff
@@ -23,15 +24,26 @@ export function extractIdFromLink(
   return segments[segments.length - 1] || "";
 }
 
+function transformResourcesName(name: string | undefined | null): ResourceType {
```
madhuravius, line 27: is this to singularize? might make sense to have a more meaningful utility name so people know, or are apps/databases unique to this

neurosnap: Resource type is a special thing inside `deploy-api` that is not sent to the client.  This is an area for improvement on the API and not something I wanted to block development of the FE.

I added a comment to describe why that function exists.

src/ui/pages/create-project-git.tsx
```diff
@@ -618,11 +650,236 @@ export const CreateProjectGitSettingsPage = () => {
...
+  return `${op.resourceType}:${op.type}`;
+};
+
+const createReadableStatus = (status: OperationStatus): string => {
+  switch (status) {
+    case "queued":
+      return "Queued";
+    case "running":
+      return "Pending";
+    case "succeeded":
```
madhuravius, line 682: stylistic oddity here, casing difference, i guess intentional?

src/ui/pages/create-project-git.tsx
```diff
@@ -618,11 +650,236 @@ export const CreateProjectGitSettingsPage = () => {
...
+    setDate(prettyDateRelative(from));
+  }, [from]);
+  useInterval(() => setDate(prettyDateRelative(from)), 1 * 1000);
+
+  const className = cn(
+    "rounded-full border-2",
+    "text-sm font-semibold ",
+    "mt-1 px-2 flex justify-between items-center w-fit",
+  );
+  // style={{ transform: "scale(0.7)" }}
```
madhuravius, line 826: delete probably?

src/ui/pages/create-project-git.tsx
```diff
@@ -618,11 +650,236 @@ export const CreateProjectGitSettingsPage = () => {
...
   return (
-    <div className="border-b-2 py-4">
-      {resource}:{op.type} - {op.status}
+    <div
+      className={cn(className, "text-indigo border-indigo-300 bg-indigo-100")}
+    >
+      <IconInfo
+        color="#4361FF"
+        className="mr-1"
+        style={{ transform: "scale(0.7)" }}
```
madhuravius, line 878: lots of repetition here might be worthy of class name + css?

src/ui/pages/create-project-git.tsx
```diff
@@ -618,11 +650,236 @@ export const CreateProjectGitSettingsPage = () => {
...
+    );
+  }
+
+  if (status === "provisioned") {
+    return (
+      <div className={cn(className, "text-forest border-lime-300 bg-lime-100")}>
+        <IconCheck
+          color="#00633F"
+          style={{ transform: "scale(0.7)" }}
+          height={20}
```
madhuravius, line 863: i swore in a different PR i saw standardized sizing of icons (does this need to be set explicitly?)

src/ui/pages/create-project-git.tsx
```diff
@@ -658,19 +918,35 @@ export const CreateProjectGitStatusPage = () => {
...
+                alt="default project logo"
+                src="/logo-app.png"
+                style={{ width: 32, height: 32 }}
+                className="mr-3"
+              />
+            </div>
+            <div>
+              <h4 className={tokens.type.h4}>{env.handle}</h4>
+              <p className="text-black-500 text-sm">
+                https://aptible.com/839583485/dashboard
```
madhuravius, line 934: intentional?

src/deploy/database/index.ts
```diff
@@ -29,7 +29,7 @@ import { createSelector } from "@reduxjs/toolkit";
 import { findEnvById, selectEnvironments } from "../environment";
 
 export interface DeployDatabaseResponse {
-  id: string;
+  id: number;
```
madhuravius, line 32: this is confusing but wanted to make sure you're aware, you're accepting it as a number and then passing it back as a string? why not just use one?

madhuravius (approved):
LGTM - minor questions/nits/things to check

Ping if reappr needed

----

Side question, for common utilities like time can we just use date-fns or a stable/popular library if we're doing date comparisons/math? I am uncomfortable with the idea of maintaining a set of utilities there if that gets expanded

src/ui/pages/create-project-git.tsx
neurosnap: Intentional @joeclayallday

src/ui/pages/create-project-git.tsx
neurosnap: Follow up PR deletes

src/ui/pages/create-project-git.tsx
neurosnap: Following PR reverts a bunch of this icon styling

src/ui/pages/create-project-git.tsx
neurosnap: We are trying to shrink the icons but it's not working well with the SVGs.  Follow up PR reverts so we can create a new PR to handle this use case properly.

src/ui/pages/create-project-git.tsx
neurosnap: Placeholder, intentional.  We don't have any endpoints yet (pending your work)

src/deploy/database/index.ts
neurosnap: Yea I need to rethink this.  I prefer ids as strings and I think we might have some code that assumes it's a string.  I'll need to revert and play around with it again.

neurosnap:
We use `date-fns` inside the `src/date` package.  All date utilities should be exported from there and then consumed by other packages.  This makes it easier for us to swap-out the date library if needed.

madhuravius:
misread the date file diff (it obscured import on cli) : https://github.com/aptible/cloud-ui/blob/main/src/date/index.ts

LGTM

PR #18, opened 2023-02-14
ftux status rely ops

neurosnap:
Screen Shot 2023-02-14 at 3 01 53 PM

https://aptible.slack.com/archives/C022BUSK487/p1676402464109179

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -42,6 +42,8 @@ import {
   IconCheck,
   IconXMark,
   IconInfo,
+  IconGitCommit,
+  IconGitBranch,
 } from "../shared";
 import { AddSSHKeyForm } from "../shared/add-ssh-key";
 import { createProject, deployProject, TextVal } from "@app/projects";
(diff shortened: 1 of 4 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -790,79 +794,102 @@ const ProjectStatus = ({
...
   const className = cn(
     "rounded-full border-2",
     "text-sm font-semibold ",
-    "mt-1 px-2 flex justify-between items-center w-fit",
+    "px-2 flex justify-between items-center w-fit",
   );
-  // style={{ transform: "scale(0.7)" }}
 
-  if (status === "pending") {
+  if (status === "running" || status === "queued") {
```
madhuravius, line 869: a nit, but for these values that seem to be enums/constants of some sort, makes sense to not check them as strings repeatedly in different files (easy for someone to make a mistake), might be worthy dumping them in a constants.ts type location and referencing them

madhuravius: disregard above, this doesn't mean anything because it comes from the OperationStatus type anyway

src/ui/pages/create-project-git.tsx
```diff
@@ -790,79 +794,102 @@ const ProjectStatus = ({
...
+
+    if (st.status === "succeeded") {
+      success += 1;
+    }
   }
-  if (stats.every((s) => s.status === "provisioned")) {
-    return "provisioned";
+
+  if (success === statuses.length) {
+    return ["succeeded", statuses[statuses.length - 1].updatedAt];
```
madhuravius, line 828: you can also do .at(-1), i think now, not sure if we have that transpiler for this

madhuravius (approved):

src/ui/pages/create-project-git.tsx
neurosnap: Oh interesting, that's good to know, thanks!

madhuravius (approved):

PR #19, opened 2023-02-16
feat(ftux): show operation logs

neurosnap:
Screen Shot 2023-02-16 at 1 54 45 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "redux-batched-actions": "^0.5.0",
     "redux-persist": "^6.0.0",
     "rome": "^11.0.0",
-    "saga-query": "^7.0.0",
+    "saga-query": "^7.1.0",
     "tailwindcss": "^3.2.4",
     "typescript": "^4.9.4",
     "vite": "^4.0.4",
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -8,15 +8,21 @@ import {
   errorHandler,
   dispatchActions,
   timer,
-  CreateActionWithPayload,
   put,
   setLoaderStart,
   setLoaderError,
   all,
   setLoaderSuccess,
-  LoaderCtx,
+  delay,
+  createKey,
 } from "saga-query";
-import type { ApiCtx, Next, PipeCtx } from "saga-query";
+import type {
+  ApiCtx,
+  Next,
+  PipeCtx,
+  LoaderCtx,
+  CreateActionWithPayload,
+} from "saga-query";
 
 import { selectEnv } from "@app/env";
 import type { ApiGen, AuthApiError, Action, HalEmbedded } from "@app/types";
@@ -159,7 +165,7 @@ export interface PaginateProps {
 }
 
 /**
- * Loops through all the pages of an endpoint
+ * Loops through all the pages of an endpoint automatically
  */
 export function combinePages<
   P extends { [key: string]: any } = { [key: string]: any },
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -59,7 +59,7 @@
     "redux-batched-actions": "^0.5.0",
     "redux-persist": "^6.0.0",
     "rome": "^11.0.0",
-    "saga-query": "^7.0.0",
+    "saga-query": "^7.1.0",
```
madhuravius, line 62: should prob dependabot this and we shouldn't be using `npm i` or equivalent `yarn install --frozen-lockfile` which i think is appropriate here (not yarn) - https://github.com/aptible/cloud-ui/blob/main/Dockerfile#LL12C9-L12C9

madhuravius: this is a nit obv, but raising for awareness

src/api/index.ts
```diff
@@ -206,6 +212,61 @@ export function combinePages<
...
 
+export interface Retryable {
+  attempts?: number;
+}
+
+/**
+ * This middleware will retry calling the endpoint when there is a failure until
+ * `maxAttempts` is reached.
+ */
+export function retryable(
```
madhuravius, line 223: huge nitpick but the file is inconsistent (we can do this later) with function notation vs const notation for functions  (which is nice in it prevents reassignment or some degree of top level modification), which would change some of these signatures (not the type signatures anyway) a little and how we read them, just noticed

src/api/index.ts
```diff
@@ -206,6 +212,61 @@ export function combinePages<
...
+
+    yield delay(waitTime);
+    const fn = ctx.actionFn;
+    // call the endpoint again
+    const response = yield* call(
+      fn.run,
+      fn({ ...ctx.payload, attempts: attempts + 1 }),
+    );
+    // check for success
+    if ((response as any).json.ok) {
```
madhuravius, line 259: probably wanna come back to the response as any? (probably referencing a specific type here that has a json property?), if it's actually any we should `?.` it probably

src/ui/pages/create-project-git.tsx
```diff
@@ -902,12 +917,86 @@ const StatusPill = ({
...
+  if (op.status === "queued" || op.status === "running") {
+    return (
+      <div className={wrapper}>
+        Operation {op.status}, logs will display after operation completes.
+      </div>
+    );
+  }
+
+  if (loader.isInitialLoading) {
+    return <div className={wrapper}>Fetching logs ...</div>;
```
madhuravius, line 984: nit/idea - hopefully this is so fast the user doesn't notice but you had a loader component that might look nicer here rather than unformatted text

src/ui/pages/create-project-git.tsx
```diff
@@ -902,12 +917,86 @@ const StatusPill = ({
...
+  }
+
+  if (!data) {
+    return <div className={wrapper}>No data found</div>;
+  }
+
+  return (
+    <div className={wrapper}>
+      {data.split("\n").map((line, i) => {
```
madhuravius, line 993: nit/choice - in arrow functions the return is optional, right?

```jsx
{data.split("\n").map((line, i) => (<LogLine key={`log-${i}`} text={line} />)))}
```

or something like that, if not i'd just use functional notation as usually linters should be mad about this 🤔

src/ui/pages/create-project-git.tsx
```diff
@@ -735,11 +734,27 @@ const Op = ({
   };
 
   return (
-    <div className={`${extra} py-4 flex justify-between items-center`}>
-      <div className="font-semibold">
-        {createReadableResourceName(op, resource)}
+    <div className={extra}>
+      <div
+        className="py-4 flex justify-between items-center cursor-pointer"
+        onClick={() => setOpen(!isOpen)}
```
madhuravius, line 740: pref, but might be more readable to do a `onClick={toggleOpen}`, which does the same thing as this with the event, and putting the shared code in the toggle

madhuravius (approved):
LGTM, nits/followup questions (not to this PR itself) but mostly not immediately relevant items

poke if reappr needed

src/api/index.ts
neurosnap: Yea this is an issue with `saga-query` typing that I need to fix.

PR #22, opened 2023-02-22
chore(ftux): use saga-query fetchRetry middleware

neurosnap:
The `fetchRetry` middleware from `saga-query` accesses the actual fetch request and will keep calling it multiple times until there is a success.

You can read more here: https://github.com/redux-saga/saga-query/releases/tag/v7.2.0


madhuravius (approved):
LGTM, seems easy enough: https://github.com/redux-saga/saga-query/pull/29

looks like backoff is implicit and exponential by default, which is nice, i should be able to use this in comparable functions

PR #23, opened 2023-02-23
feat: initial attempt at service definitions in flow

madhuravius:
Lots of things happening here, but it LOOKS to work properly:

image

image

This is very much a brute force approach (it will destroy all service definitions if found for an app (if you hit the back button for example or try to fix), and recreate them). it's quite quick but i suspect this will result in suboptimal behavior (we prob wanna use the updates where possible, but that's harder to do without the new component)

image

---

This works as well:

image

### DIFF
--- a/src/deploy/app-service-definitions/index.ts
+++ b/src/deploy/app-service-definitions/index.ts
@@ -1,0 +1,116 @@
+import { createSelector } from "@reduxjs/toolkit";
+import {
+  api,
+  cacheTimer,
+  combinePages,
+  DeployApiCtx,
+  PaginateProps,
+  thunks,
+} from "@app/api";
+import { defaultEntity, extractIdFromLink } from "@app/hal";
+import {
+  createReducerMap,
+  createTable,
+  mustSelectEntity,
+} from "@app/slice-helpers";
+import type {
+  AppState,
+  DeployServiceDefinition,
+  LinkResponse,
+} from "@app/types";
+import { selectDeploy } from "../slice";
+
+export interface DeployServiceDefinitionResponse {
+  id: string;
+  command: string;
+  process_type: string;
+  created_at: string;
+  updated_at: string;
+  _links: {
+    app: LinkResponse;
+  };
+}
+
+export const defaultDeployServiceDefinition = (
+  e: Partial<DeployServiceDefinition> = {},
+): DeployServiceDefinition => {
+  const now = new Date().toISOString();
+  return {
+    createdAt: now,
+    updatedAt: now,
+    appId: "",
+    command: "",
+    processType: "",
+    ...e,
+  };
+};
+
+export const deserializeServiceDefinition = (
+  payload: DeployServiceDefinitionResponse,
+): DeployServiceDefinition => {
+  const links = payload._links;
+  return {
+    appId: extractIdFromLink(links.app),
+    command: payload.command,
+    processType: payload.process_type,
+    createdAt: payload.created_at,
+    updatedAt: payload.updated_at,
+  };
+};
+
+export const DEPLOY_SERVICE_DEFINITION_NAME = "serviceDefinitions";
+const slice = createTable<DeployServiceDefinition>({
+  name: DEPLOY_SERVICE_DEFINITION_NAME,
+});
+const { add: addServiceDefinitions } = slice.actions;
+export const serviceDefinitionReducers = createReducerMap(slice);
+
+const selectors = slice.getSelectors(
+  (s: AppState) => selectDeploy(s)[DEPLOY_SERVICE_DEFINITION_NAME],
+);
+export const { selectTableAsList: selectAppsAsList } = selectors;
+const initApp = defaultDeployServiceDefinition();
+const must = mustSelectEntity(initApp);
+export const selectServiceDefinitionById = must(selectors.selectById);
+
+export const fetchServiceDefinitionsByAppId = api.get<{ appId: string }>(
+  "/apps/:appId/service_definitions",
+  { saga: cacheTimer() },
+  api.cache(),
+);
+
+export const deleteServiceDefinition = api.delete<{
+  serviceDefinitionId: string;
+}>("/service_definitions/:serviceDefinitionId");
+
+export const fetchServiceDefinition = api.get<{ serviceDefinitionId: string }>(
+  "/service_definitions/:serviceDefinitionId",
+);
+
+interface CreateServiceDefinitionProps {
+  appId: string;
+  command: string;
+  processType: string;
+}
+
+export const createServiceDefinition = api.post<
+  CreateServiceDefinitionProps,
+  DeployServiceDefinitionResponse
+>("/apps/:appId/service_definitions", function* (ctx, next) {
+  const { command, processType } = ctx.payload;
+  const body = {
+    command: command,
+    process_type: processType,
+  };
+  ctx.request = ctx.req({ body: JSON.stringify(body) });
+
+  yield next();
+});
+
+export const serviceDefinitionEntities = {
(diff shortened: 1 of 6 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -454,9 +466,38 @@ export const CreateProjectGitSettingsPage = () => {
...
   );
+  const [existingCmds, setExistingCmds] = useState<TextVal[]>([]);
+  const cmdList = parseText(cmds);
+
   const loader = useSelector((s: AppState) =>
     selectLoaderById(s, { id: `${deployProject}` }),
   );
+
+  useEffect(() => {
+    const serviceDefinitions =
```
madhuravius, line 477: i have to rewrite this in the spirit of how you currently write some of your loops (sorry for taking func paradigm on this, just written in haste)

src/ui/pages/create-project-git.tsx
```diff
@@ -498,7 +539,8 @@ export const CreateProjectGitSettingsPage = () => {
         envId: app.environmentId,
         dbs: dbList,
         envs: envList,
-        cmds: [],
+        cmds: cmdList,
+        existingCmds,
```
madhuravius, line 543: should drop this and keep only one list (can't do that until we switch out of text area as that's an entropic input type right now)

src/ui/pages/create-project-git.tsx
```diff
@@ -442,9 +450,13 @@ export const CreateProjectGitSettingsPage = () => {
   const app = useSelector((s: AppState) => selectAppById(s, { id: appId }));
   const { scanOp, codeScan, appOps } = useLatestCodeResults(appId);
 
-  const query = useQuery(fetchAllDatabaseImages());
+  const dbQuery = useQuery(fetchAllDatabaseImages());
   const dbImages = useSelector(selectDatabaseImagesAsList);
 
+  const serviceDefinitionsQuery = useCache<HalEmbedded<HalServiceDefinition>>(
```
madhuravius, line 456: while this works we really should probably usequery here => and access via selector. i was able to fire the call but unable to persist it in state [as this](https://github.com/aptible/cloud-ui/pull/23/files#diff-c85676d450531ae3bb2c0332ef7006ddb31539238617c6bbf4b08653b0e6162fR76) when written like the other dispatched async actions (or equivalent naming), don't seem to deserialize arrays properly

src/deploy/app-service-definitions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  command: string;
+  process_type: string;
+  created_at: string;
+  updated_at: string;
+  _links: {
+    app: LinkResponse;
+  };
+}
+
+export const defaultDeployServiceDefinition = (
```
neurosnap, line 34: https://bower.sh/entity-factories

src/deploy/app-service-definitions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  name: DEPLOY_SERVICE_DEFINITION_NAME,
+});
+const { add: addServiceDefinitions } = slice.actions;
+export const serviceDefinitionReducers = createReducerMap(slice);
+
+const selectors = slice.getSelectors(
+  (s: AppState) => selectDeploy(s)[DEPLOY_SERVICE_DEFINITION_NAME],
+);
+export const { selectTableAsList: selectAppsAsList } = selectors;
+const initApp = defaultDeployServiceDefinition();
```
neurosnap, line 72: https://bower.sh/death-by-thousand-existential-checks

src/deploy/app-service-definitions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+const must = mustSelectEntity(initApp);
+export const selectServiceDefinitionById = must(selectors.selectById);
+
+export const fetchServiceDefinitionsByAppId = api.get<{ appId: string }>(
+  "/apps/:appId/service_definitions",
+  { saga: cacheTimer() },
+  api.cache(),
```
neurosnap, line 79: You don't need this anymore since everything is getting stored inside a specific redux slice.  This is only for `auto-caching` which is used when you don't have a specific redux slice setup.

src/deploy/app-service-definitions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  const body = {
+    command: command,
+    process_type: processType,
+  };
+  ctx.request = ctx.req({ body: JSON.stringify(body) });
+
+  yield next();
+});
+
+export const serviceDefinitionEntities = {
```
neurosnap, line 110: This pipes into `src/hal/index` and does auto-ETL from `deploy-api`

src/projects/index.ts
```diff
@@ -10,7 +10,7 @@ import {
 } from "saga-query";
 
 import { createLog } from "@app/debug";
-import { ThunkCtx, thunks } from "@app/api";
+import { combinePages, ThunkCtx, thunks } from "@app/api";
```
neurosnap, line 13: `combinePages` is unused I think here?

src/deploy/app-service-definitions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export const fetchServiceDefinitionsByAppId = api.get<{ appId: string }>(
+  "/apps/:appId/service_definitions",
+  { saga: cacheTimer() },
+  api.cache(),
+);
+
+export const deleteServiceDefinition = api.delete<{
+  serviceDefinitionId: string;
+}>("/service_definitions/:serviceDefinitionId");
```
neurosnap, line 84: `:id` seems fine for CRUD, we know what the id is here.

src/deploy/app-service-definitions/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export const fetchServiceDefinition = api.get<{ serviceDefinitionId: string }>(
+  "/service_definitions/:serviceDefinitionId",
```
neurosnap, line 87: same here

src/projects/index.ts
```diff
@@ -144,6 +149,37 @@ export const deployProject = thunks.create<CreateProjectSettingsProps>(
...
+          deleteServiceDefinition({
+            serviceDefinitionId: cmdId,
+          }),
+        ),
+      ),
+    );
+
+    // TODO - convert this to a series of updates where possible (currently information is agnostic)
+    // create all the new ones
+    yield all(
```
neurosnap, line 170: Could this go inside the other `all()` on line 183?

src/ui/pages/create-project-git.tsx
neurosnap: Switch it to `useQuery` and then use a selector.  You have everything setup to support it.

neurosnap (commented):
Very nice job, you caught on quick to how this system works.  It's a little magical so I'm happy to review with you

src/projects/index.ts
madhuravius: it was intended with the action to be queried on, i removed that as it wasn't working for me and i went back to using a component state based approach (just so we can test this thing out)

src/projects/index.ts
madhuravius: i don't think so, you probably want these records created before requesting to depoy an app

madhuravius:
> Very nice job, you caught on quick to how this system works. It's a little magical so I'm happy to review with you

I will probably consolidate some of this into an internal doc while I'm writing runbooks so people can add to this pretty easily with easier onboarding (FAQ/read this if curious on how to do FE on some of this)

neurosnap (approved):
Strong work!

neurosnap (approved):

PR #24, opened 2023-02-23
feat(ftux): show back button on status page when there's a failure

neurosnap:
We also allow the user to make changes to environment variables and service commands without re-provisioning databases.

We also require user to provide a database handle instead of generating one for them.

https://user-images.githubusercontent.com/1940365/221024042-44da351e-be69-45f0-8056-9aa0a87e9df2.mp4

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -379,12 +379,12 @@ const validateDbs = (
         message: `[${item.key}] is not a valid handle: /\A[0-9a-z._-]{1,64}\z/`,
       });
     }
-    const [type = "", version = ""] = item.value.split(":");
-    const imgs = dbImages.filter((img) => img.type === type);
+    const [_type = "", version = ""] = item.value.split(":");
+    const imgs = dbImages.filter((img) => img.type === _type);
     if (imgs.length === 0) {
       errors.push({
         item,
-        message: `[${type}] is not a valid database`,
+        message: `[${_type}] is not a valid database`,
       });
       return;
     }
@@ -399,7 +399,7 @@ const validateDbs = (
     if (!found) {
       errors.push({
         item,
-        message: `[${version}] is not a valid version for [${type}]`,
+        message: `[${version}] is not a valid version for [${_type}]`,
       });
     }
   };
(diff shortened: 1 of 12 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -366,26 +373,33 @@ const validateDbs = (
...
 
   const validate = (item: TextVal) => {
-    const imgs = dbImages.filter((img) => img.type === item.key);
+    if (!/^[0-9a-z._-]{1,64}$/.test(item.key)) {
+      errors.push({
+        item,
+        message: `[${item.key}] is not a valid handle: /\A[0-9a-z._-]{1,64}\z/`,
+      });
+    }
+    const [type = "", version = ""] = item.value.split(":");
```
madhuravius, line 382: nit but might be useful to not use a reserved word `type`

src/ui/pages/create-project-git.tsx
```diff
@@ -555,41 +603,50 @@ export const CreateProjectGitSettingsPage = () => {
...
+                {existingEnvStr ? (
+                  <p>
+                    Databases have already been created so you cannot make
+                    changes to them in this screen anymore.
+                  </p>
+                ) : (
+                  <>
+                    <p>
+                      You can provide as many databases as you need (each line
+                      is a separate database, e.g. NAME=DB:VERSION).
```
madhuravius, line 621: might wanna put that `NAME=DB:VERSION` in a pre tag

src/ui/pages/create-project-git.tsx
```diff
@@ -619,15 +673,19 @@ export const CreateProjectGitSettingsPage = () => {
             label="Service and Commands"
             htmlFor="commands"
             feedbackVariant="info"
+            description={
+              <div>
+                <p>
+                  Each line is separated by a service command in format:
```
madhuravius, line 679: same as above for the code/formatted text

src/ui/pages/create-project-git.tsx
neurosnap: @joeclayallday said to remove those

src/ui/pages/create-project-git.tsx
neurosnap: @joeclayallday said to remove those

madhuravius (approved):

PR #25, opened 2023-02-27
webauthn login

neurosnap:
This feature makes it possible to login using a security instead of OTP.

This shows the flow, there's an error because it's really hard to test locally.

https://user-images.githubusercontent.com/1940365/221666932-91a3755b-56de-49e2-b30d-5cc1df1ea493.mov

### DIFF
--- a/src/auth/webauthn.ts
+++ b/src/auth/webauthn.ts
@@ -1,0 +1,42 @@
+import type {
+  CredentialCreationOptionsJSON,
+  CredentialRequestOptionsJSON,
+} from "@github/webauthn-json";
+import { get, create } from "@github/webauthn-json";
+
+export function hasWebAuthnSupport(): boolean {
+  let scope = null;
+  if (window) {
+    scope = window;
+  } else if (global) {
+    scope = global;
+  } else {
+    return false;
+  }
+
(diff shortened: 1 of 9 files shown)


src/auth/webauthn.ts
```diff
@@ -0,0 +1,42 @@
...
+export function webauthnCreate(payload: CredentialCreationOptionsJSON) {
+  if (!hasWebAuthnSupport()) {
+    throw new Error("webauthn not supported for browser");
+  }
+
+  return create(payload);
+}
+
+export function webauthnGet(
+  payload: CredentialRequestOptionsJSON["publicKey"] & { challenge: string },
```
madhuravius, line 30: interesting notation, why not just `CredentialRequestOptionsJSON.publicKey`? (link to typ)[https://github.com/github/webauthn-json/blob/main/src/webauthn-json/basic/json.ts#L83]

threw me off

madhuravius (approved):
nit that may be worthy of following just for cleanliness, but LGTM

ping if reappr required

src/auth/webauthn.ts
neurosnap: For some reason they didn't export that type and I believe TS requires this notation.

PR #26, opened 2023-02-28
feat: initial attempt at environments list and environment tabs

madhuravius:
Latest image:

image

---

Older images for ref:

List view (button is disabled as it is aspirational):

image

And populated tabs (empty for now):

image

image

etc.

Lots of todos still outstanding:

- [x] environment activity is pending / todo, need more info on what that means
- [x] operations are pretty wrong (last retrieved is not really sorted properly), similarly statuses are wrong
- [x] not sure what create environment means in our current world (probably other tickets?)
- [x] there is some debate/discussion as to what should be shown but I'm unsure if it should be db/app/etc counts, or statuses (or whatever is meaningful)
- [x] not 100% sure what those options entail, will try to match (but need some info):

image


joeclayallday:
We can make those • • • Action buttons smaller too
Screen Shot 2023-02-27 at 8 38 21 PM
Screen Shot 2023-02-27 at 8 38 14 PM
Screen Shot 2023-02-27 at 8 38 28 PM

madhuravius:
Current state after changes requested / being closer to Figma (some minor spacing/centering issues still outstanding):

image

src/date/index.ts
```diff
@@ -1,4 +1,55 @@
...
+      return `in ${result}`;
+    } else {
+      return `${result} ago`;
+    }
+  }
+
+  return result;
+};
+
+// taken from
```
madhuravius, line 39: correctly reference it

src/ui/shared/environment/environment-activity.tsx
```diff
@@ -0,0 +1,71 @@
...
+  } else if (operationStatus === "queued" || operationStatus === "running") {
+    return <IconInfo color="#4361FF" />;
+  }
+  return <IconXCircle color="#AD1A1A" />;
+};
+
+export const EnvironmentActivity = () => {
+  const operations = useSelector((s: AppState) =>
+    selectOperationsAsList(s),
+  ).slice(0, 5);
```
madhuravius, line 23: what # do we want here?

neurosnap: 5 seems good to me for now.  If you move this to its own selector then it'll be better memoized.  Non-blocker.

src/ui/shared/environment/environment-activity.tsx
```diff
@@ -0,0 +1,71 @@
...
+                <tr>
+                  <th scope="col" className="text-left px-3 py-4" colSpan={2}>
+                    <span className={tokens.type["small normal lighter"]}>
+                      Recent Activity
+                    </span>
+                  </th>
+                </tr>
+              </thead>
+              <tbody className="divide-y divide-gray-200 bg-white">
+                {operations.map((operation) => (
```
neurosnap, line 42: If you move this into its own `EnvironmentOperationView` component then you can use `selectEnvironmentById` which will be less gnarly than `selectEnvironments` (which will rerender this component everytime there's a change to this slice).

src/ui/shared/environment/environment-activity.tsx
```diff
@@ -0,0 +1,71 @@
...
+                      {IconForResource(operation.status)}
+                    </Td>
+                    <Td className="pl-0 2xl:flex-cell-md sm:flex-cell-sm">
+                      <span className="font-semibold text-black">
+                        {capitalize(operation.resourceType)} {operation.type}{" "}
+                        {operation.status}{" "}
+                      </span>
+                      <span>
+                        for{" "}
+                        {environments?.[operation.environmentId]?.handle ??
```
neurosnap, line 54: 💀 we should be able to remove this lookup with comment above.

src/ui/shared/environment/environment-list.tsx
```diff
@@ -0,0 +1,203 @@
...
+  return (
+    <Td className="2xl:flex-cell-md sm:flex-cell-sm">
+      <div
+        className={tokens.type.darker}
+        style={{ textTransform: "capitalize" }}
+      >
+        {prettyEnglishDate(operation.createdAt)}
+      </div>
+      <div>
+        {timeAgo(operation.createdAt)} by {operation.userName || "Unknown"}
```
neurosnap, line 123: We can make the default value for `userName` be `Unknown` that way you don't need to do this `||` check here.

src/ui/shared/tokens.ts
```diff
@@ -15,9 +15,12 @@ export const tokens = {
     "subdued active link": "font-medium text-sm text-emerald-600",
 
     "small semibold darker": "text-sm font-semibold text-gray-900",
+    "small normal darker": "text-sm font-normal text-gray-900",
```
neurosnap, line 18: I've been trying to get rid of these actually.  Not sure how you feel about them but they feel a little awkward.

src/deploy/app/index.ts
```diff
@@ -114,6 +119,34 @@ export interface DeployAppRow extends DeployApp {
   envHandle: string;
 }
 
+const selectSearchProp = (_: AppState, props: { search: string }) =>
+  props.search.toLocaleLowerCase();
+
+export const selectEnvironmentsForTableSearch = createSelector(
```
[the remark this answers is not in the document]
madhuravius: yes, mistake

src/ui/shared/environment/environment-activity.tsx
madhuravius: i think i also have a mistake somewhere as i'm not calling the action beforehand (so my operation list is empty)

src/ui/shared/tokens.ts
madhuravius: i think we should remove them, it's super awkward when we can just do it the tailwind way without the helper intermediate

src/deploy/app/index.ts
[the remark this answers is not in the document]
madhuravius: realized why i did it (shared import to lead to wonky non self-enclosed imports) as i reference the selectSearchProp in the new func (which is in itself a new func)

I'll move it but we should be mindful of multiple chains of imports (not that it matters too much nowadays except hygiene)

src/ui/pages/environments.tsx
```diff
@@ -0,0 +1,37 @@
...
+import {
+  ButtonIcon,
+  EnvironmentActivity,
+  EnvironmentList,
+  IconPlusCircle,
+  TableHead,
+} from "../shared";
+
+export const EnvironmentsPage = () => {
+  // TODO: BELOW BUTTON SUCKS, FIND A WAY TO LINE IT UP
```
madhuravius, line 12: any thoughts on this? i was initially debating moving this right bar into a pair of the activity table + list table (with the headings + table header pulled out into another component)

this still feels gnarly, leaving it in and maybe we can refactor this when we add this to every page?

src/ui/shared/environment/environment-activity.tsx
madhuravius: I moved it, but seems weird 

image

src/ui/shared/tokens.ts
madhuravius: sooner is better than later, probably best to rip the bandaid off, can do so if you wish:

image

src/ui/shared/environment/environment-activity.tsx
madhuravius: i actually initially meant to do this for all environments based on the sidebar misleading me (i think it was an environment create with OLDER operations behind it, implying multiple envs reflect on the activity panel)

i will adjust this and it should be cleaner + slightly perf improvement (not really notable yet)

madhuravius: i lied, part of my previous statement is correct (this is for all envs), but i will still move this into its own component with an environment passed in to avoid rerendering the entire component

madhuravius:
I had an item in my bullet list we should just discuss when you get a min, but i think we need to hydrate the operations list somewhere (i don't seem to have them all in redux). original `operations are pretty wrong (last retrieved is not really sorted properly), similarly statuses are wrong` in todos for this pr

src/ui/shared/environment/environment-activity.tsx
neurosnap: To be clear: I would create a new component where you only pass in the operation.  Then inside the component you do a lookup for the environment.

madhuravius: https://github.com/aptible/cloud-ui/pull/26/commits/928644cdf3de45a9b2beeffd5467a241dbe90858#diff-a2f31509bc43ca57054509c5656c7f45cbd4355988dc80c3e64db3a3131a46c2R43 <-- that's what I did i think

neurosnap (dismissed):
https://user-images.githubusercontent.com/1940365/222192838-86dbaeb4-8a04-4a89-a1cf-006d1550122d.mov

neurosnap (approved):

PR #28, opened 2023-03-01
chore: added redux dev tools extension optionally

madhuravius:
Leaving as draft. This should have no impact (or nearly no impact). There are all sorts of blog posts and stuff about customizing this thing for prod: https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f, but I think we can really just use this and there's some primitive settings we can use to disable in prod if you wish.

Similarly we're using a deprecated method, which I'm trying to solve here:

image

---

Proof:

image

Pretty handy dandy to view state:

image

### DIFF
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -1,4 +1,4 @@
-import { createStore, applyMiddleware } from "@reduxjs/toolkit";
+import { createStore, applyMiddleware, compose } from "@reduxjs/toolkit";
 import type { Middleware, Store } from "@reduxjs/toolkit";
 import { PersistPartial } from "redux-persist/es/persistReducer";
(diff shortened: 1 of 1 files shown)


madhuravius:
After changes (still seems to work?)

image

src/app/store.ts
```diff
@@ -1,6 +1,5 @@
-import { createStore, applyMiddleware } from "@reduxjs/toolkit";
+import { compose, configureStore } from "@reduxjs/toolkit";
```
neurosnap, line 1: `compose` 💀

neurosnap (approved):

src/app/store.ts
madhuravius: dropped my bad

PR #29, opened 2023-03-02
feat(ftux): create default endpoint after app deploys successfully

neurosnap:
https://user-images.githubusercontent.com/1940365/223149104-91b79431-2b91-47c0-9029-2400f3dfb240.mp4

### DIFF
--- a/src/deploy/endpoint/index.ts
+++ b/src/deploy/endpoint/index.ts
@@ -138,20 +138,15 @@ export const endpointEntities = {
 };
 
 export interface CreateEndpointProps {
-  appId: string;
   serviceId: string;
 }
 
 export const createEndpoint = api.post<CreateEndpointProps>(
   "/services/:serviceId/vhosts",
   function* (ctx, next) {
-    const { appId, serviceId } = ctx.payload;
     const body = JSON.stringify({
-      app_id: appId,
-      service_id: serviceId,
       platform: "alb",
       type: "http_proxy_protocol",
-      status: "pending",
       default: true,
       acme: false,
       internal: false,
--- a/src/deploy/endpoint/index.ts
+++ b/src/deploy/endpoint/index.ts
@@ -1,13 +1,28 @@
-import { api, cacheTimer } from "@app/api";
+import {
+  call,
+  put,
+  setLoaderError,
+  setLoaderStart,
+  setLoaderSuccess,
+} from "saga-query";
+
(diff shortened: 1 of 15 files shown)


src/projects/index.ts
```diff
@@ -148,77 +177,177 @@ export const deployProject = thunks.create<CreateProjectSettingsProps>(
...
-        createAppOperation({
-          type: "configure",
-          appId,
-          env: envs.reduce(
-            (acc, env) => ({ ...acc, [env.key]: env.value }),
-            {},
-          ),
-        }),
-      ),
+    const env: { [key: string]: any } = {};
```
madhuravius, line 196: our backend supports anything besides string/integer for the env value (?). might wanna check this (we might only support string values, or string/int only)

src/deploy/endpoint/index.ts
```diff
@@ -114,3 +136,106 @@ export const endpointEntities = {
...
+    const body = JSON.stringify({
+      app_id: appId,
+      service_id: serviceId,
+      platform: "alb",
+      type: "http_proxy_protocol",
+      status: "pending",
+      default: true,
+      acme: false,
+      internal: false,
+    });
```
joshraker, line 158: There's a few extra parameters here: `app_id`, `service_id` (pulled from the path), and `status`. They'll just be ignored so it's not a big deal unless we do start using them for some reason in the future. You can find the list of parameters expected on a create [here](https://github.com/aptible/deploy-api/blob/master/app/controllers/vhosts_controller.rb#L45-L49).

src/ui/pages/create-project-git.tsx
```diff
@@ -426,6 +432,7 @@ const validateEnvs = (items: TextVal[]): ValidatorError[] => {
     }
   };
 
+  console.log(items);
```
madhuravius, line 435: might wanna drop

src/ui/pages/create-project-git.tsx
```diff
@@ -443,10 +450,190 @@ const useLatestCodeResults = (appId: string) => {
...
+};
+
+const idCreator = () => {
+  let id = 0;
+  return () => {
+    id += 1;
+    return id;
+  };
+};
+const createId = idCreator();
```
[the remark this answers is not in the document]
madhuravius: this is smarter than me, ignore

madhuravius (approved):

PR #30, opened 2023-03-03
feat: added settings/deprovision page for apps

madhuravius:
* This adds a common component that is sort-of reused in both app/database pages
* This allows for deprovisioning of an app/database (and it looks to work!, though I need to check with Josh on some of it as some operations failed with rollback failed)
* Cleans up some tab behavior (I will likely need to talk to you about this Eric as this alters the index behavior)

Edit - can update to include screenshots but I think e2e testing or something deprovisioned my testing stack and I'll probably test against it in the morn for a fresh batch. It looks pretty close to the current mocks though! (Screenshot a bit below but that's slightly outdated)

app management:
image

db management:
image

### DIFF
--- a/src/app/router.tsx
+++ b/src/app/router.tsx
@@ -69,12 +69,12 @@ const appRoutes: RouteObject[] = [
             index: true,
             element: <AppsPage />,
           },
+
           {
-            path: routes.APP_DETAIL_PATH,
             element: <AppDetailLayout />,
             children: [
               {
-                index: true,
+                path: routes.APP_OVERVIEW_PATH,
                 element: <AppDetailPage />,
               },
               {
--- a/src/app/router.tsx
+++ b/src/app/router.tsx
@@ -103,11 +103,10 @@ const appRoutes: RouteObject[] = [
           },
 
           {
-            path: routes.DATABASE_DETAIL_PATH,
             element: <DatabaseDetailLayout />,
             children: [
               {
-                index: true,
+                path: routes.DATABASE_OVERVIEW_PATH,
                 element: <DatabaseDetailPage />,
               },
               {
--- a/src/projects/index.ts
+++ b/src/projects/index.ts
@@ -13,6 +13,7 @@ import { createLog } from "@app/debug";
 import { ThunkCtx, thunks } from "@app/api";
 import {
   createAppOperation,
+  createDatabaseOperation,
   createDeployApp,
   createDeployEnvironment,
   fetchDatabase,
@@ -21,6 +22,7 @@ import {
   hasDeployEnvironment,
   provisionDatabase,
   selectAppById,
+  selectDatabaseById,
   selectEnvironmentByName,
 } from "@app/deploy";
 import { createServiceDefinition } from "@app/deploy/app-service-definitions";
@@ -363,6 +365,24 @@ export const deprovisionApp = thunks.create<{
     createAppOperation({
       type: "deprovision",
       appId,
+    }),
+  );
+
+  if (!deprovisionCtx.json.ok) return;
+  yield* call(waitForOperation, { id: `${deprovisionCtx.json.data.id}` });
+});
+
+export const deprovisionDatabase = thunks.create<{
+  dbId: string;
+}>("deprovision-database", function* (ctx, next) {
+  const { dbId } = ctx.payload;
+  yield* select(selectDatabaseById, { id: dbId });
+
+  const deprovisionCtx = yield* call(
+    createDatabaseOperation.run,
+    createDatabaseOperation({
+      type: "deprovision",
+      dbId,
     }),
   );
 
--- a/src/projects/index.ts
+++ b/src/projects/index.ts
@@ -351,3 +351,21 @@ export const updateEnvWithDbUrls = thunks.create<{
 
   yield next();
 });
+
+export const deprovisionApp = thunks.create<{
+  appId: string;
+}>("deprovision-app", function* (ctx, next) {
+  const { appId } = ctx.payload;
+  yield* select(selectAppById, { id: appId });
+
+  const deprovisionCtx = yield* call(
+    createAppOperation.run,
+    createAppOperation({
+      type: "deprovision",
+      appId,
+    }),
+  );
+
+  if (!deprovisionCtx.json.ok) return;
+  yield* call(waitForOperation, { id: `${deprovisionCtx.json.data.id}` });
+});
--- a/src/projects/index.ts
+++ b/src/projects/index.ts
@@ -353,39 +353,3 @@ export const updateEnvWithDbUrls = thunks.create<{
 
   yield next();
 });
-
-export const deprovisionApp = thunks.create<{
-  appId: string;
-}>("deprovision-app", function* (ctx, next) {
-  const { appId } = ctx.payload;
-  yield* select(selectAppById, { id: appId });
-
-  const deprovisionCtx = yield* call(
-    createAppOperation.run,
-    createAppOperation({
-      type: "deprovision",
-      appId,
-    }),
-  );
-
-  if (!deprovisionCtx.json.ok) return;
-  yield* call(waitForOperation, { id: `${deprovisionCtx.json.data.id}` });
-});
-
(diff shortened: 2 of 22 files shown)


madhuravius:
03/22 - Looks a bit closer to this, need some advice from Joe on colors (I opted for `amber-COLORCODE` from tailwind, but we might want something more specific)

image

Outstanding items in todos as I resumed this work:

- font sizes, will match figma
- reskin tab colors
- make sure things are full width on responsive
- for docs links just point to main docs pages
- breadcrumbs
- text color differences in precode

madhuravius:
I also need to finish wrapping up how deprovision works (it should actually work/finish)

src/app/router.tsx
```diff
@@ -68,12 +68,12 @@ const appRoutes: RouteObject[] = [
...
             element: <AppsPage />,
           },
+
           {
-            path: routes.APP_DETAIL_PATH,
             element: <AppDetailLayout />,
             children: [
               {
-                index: true,
+                path: routes.APP_OVERVIEW_PATH,
```
madhuravius, line 76: did this because we target the active url for tabs which will collide if index url is used (causing multiple tabs to be highlighted). this just seemed easier/simpler to use with an explicit path. if you strongly disagree happy to revert if we have another way to deal with the url matching (if it stays simple, just felt this was easier)

src/ui/shared/tokens.ts
```diff
@@ -12,7 +12,7 @@ export const tokens = {
     link: "font-medium text-sm text-gray-500 hover:text-gray-700",
 
     "active link": "font-medium text-sm text-gray-700",
-    "subdued active link": "font-medium text-sm text-emerald-600",
+    "subdued active link": "font-medium text-sm text-amber-600",
```
madhuravius, line 15: need to revisit amber usage

src/ui/shared/tabs.tsx
```diff
@@ -22,7 +22,7 @@ const navLink = ({ isActive }: { isActive: boolean }) =>
       [tokens.type.link]: !isActive,
       "border-transparent  hover:border-gray-300": "!isActive",
       [tokens.type["subdued active link"]]: isActive,
-      "border-emerald-500": isActive,
+      "border-amber-500": isActive,
```
madhuravius, line 25: need to revisit amber usage

src/ui/shared/tokens.ts
madhuravius: still todo (both ambers)

src/projects/index.ts
```diff
@@ -351,3 +353,39 @@ export const updateEnvWithDbUrls = thunks.create<{
...
+  const deprovisionCtx = yield* call(
+    createDatabaseOperation.run,
+    createDatabaseOperation({
+      type: "deprovision",
+      dbId,
+    }),
+  );
+
+  if (!deprovisionCtx.json.ok) return;
+  yield* call(waitForOperation, { id: `${deprovisionCtx.json.data.id}` });
```
madhuravius, line 390: TODO - create a "loader" for this and look for `useLoaderSuccess` hook for this in the component to know when to redirect (then requery the resource, if it finds nothing the redirect out)

src/projects/index.ts
```diff
@@ -351,3 +353,39 @@ export const updateEnvWithDbUrls = thunks.create<{
 
   yield next();
 });
+
+export const deprovisionApp = thunks.create<{
```
madhuravius, line 357: move this to deploy/app and deploy/database

madhuravius:
on tab selector - color of text (active needs to be the primary/black) color needs to match (thickness of line also needs to be adjusted)

src/projects/index.ts
madhuravius: doing in subsequent kickout PRs

madhuravius:
precode looks closer to this now:

image

src/ui/pages/app-detail-settings.tsx
```diff
@@ -1,3 +1,219 @@
...
+  const [isDeprovisioning, setIsDeprovisioning] = useState<boolean>(false);
+  const dispatch = useDispatch();
+
+  const { id = "" } = useParams();
+  useQuery(fetchApp({ id }));
+  const app = useSelector((s: AppState) => selectAppById(s, { id }));
+
+  useEffect(() => {
+    setHandle(app.handle);
+  }, [app?.id]);
```
neurosnap, line 34: app is guaranteed here so no need for the `?`.

src/ui/pages/app-detail-settings.tsx
```diff
@@ -1,3 +1,219 @@
...
+            {app.id} <IconCopy className="inline h-4" color="#888C90" />
+          </p>
+          <div className="mt-4 flex" />
+          <FormGroup label="" htmlFor="thumbnail">
+            <Label className="text-base font-semibold text-gray-900 block">
+              Thumbnail Image
+            </Label>
+            <div className="flex justify-between items-center">
+              <select
+                onChange={() => {}}
```
neurosnap, line 102: What is this for?

src/ui/pages/app-detail-settings.tsx
```diff
@@ -1,3 +1,219 @@
...
+            />
+          </FormGroup>
+          <Label className="text-base mt-4 font-semibold text-gray-900 block">
+            App ID
+          </Label>
+          <p>
+            {app.id} <IconCopy className="inline h-4" color="#888C90" />
+          </p>
+          <div className="mt-4 flex" />
+          <FormGroup label="" htmlFor="thumbnail">
```
neurosnap, line 96: I'm curious, why is this label empty?

src/ui/pages/app-detail-settings.tsx
```diff
@@ -1,3 +1,219 @@
...
+                placeholder="select"
+                disabled
+              >
+                <option value="test" disabled>
+                  Some Icon
+                </option>
+              </select>
+            </div>
+          </FormGroup>
+          <FormGroup label="" htmlFor="environment-variables">
```
neurosnap, line 114: same here

src/ui/pages/db-detail-settings.tsx
```diff
@@ -1,3 +1,155 @@
...
+  const [deleteConfirm, setDeleteConfirm] = useState<string>("");
+  const [handle, setHandle] = useState<string>("");
+  const [isDeprovisioning, setIsDeprovisioning] = useState<boolean>(false);
+
+  const dispatch = useDispatch();
+  const database = useSelector((s: AppState) => selectDatabaseById(s, { id }));
+
+  useEffect(() => {
+    setHandle(database.handle);
+  }, [database?.id]);
```
neurosnap, line 29: database is guaranteed.

neurosnap (commented):
overall just a couple of comments, nice job!

src/ui/pages/app-detail-settings.tsx
madhuravius: image

it's a required field, even though you can supply your own label child 

i can make it an optional param or maybe a mutually exclusive (not sure if this is possible) with a child that is a label

src/ui/pages/app-detail-settings.tsx
madhuravius: can drop it, for sure, left it in as a clear place to populate the func (instead of people doing goofy onClick for `option`, which I don't like)

src/ui/pages/app-detail-settings.tsx
neurosnap: Why did you need to supply your own label child when the FormGroup provides one?

madhuravius: there's a few options here, right?

1. allow child arguments so that the FormGroup acts as a wrapper <-- what I did and not really the best option
2. allow a property on FormGroup to let you customize styling of a label
3. allow a property on FormGroup that lets you pass in a label to use

neurosnap: I think it makes sense to keep all Form labels the same so I feel like whatever style we want for them should be consistent which means using the `label` string.

neurosnap (approved):

PR #32, opened 2023-03-24
feat: added basic error boundary

madhuravius:
Probably need to chat on the following:

* ~back button? or some other call to action~
* ~need to discuss realistic prospects for what the error message itself can be~
* ~lacking the common background image~
* this is a bit weird because I added an error component to the error route and i'm forcing it to error (we will want to skip this error for sentry if we opt to use it). I like this because we can rapidly test that the error boundary itself is stable/works, but it feels weird

---

Seems to work:

image


src/ui/shared/error-boundary.tsx
```diff
@@ -0,0 +1,75 @@
...
+            </div>
+            <div className="mt-8 sm:mx-auto sm:w-full sm:max-w-md">
+              <div className="bg-white py-8 px-4 shadow sm:rounded-lg sm:px-10">
+                <h1 className="text-lg font-semibold text-red">
+                  <IconAlertTriangle
+                    className="inline -mt-1 mr-2"
+                    color="#AD1A1A"
+                    style={{ width: 21, height: 21 }}
+                  />
+                  Error - Something went wrong
```
madhuravius, line 63: change this to `this.state.name` or default to this message

madhuravius:
@neurosnap  - https://reactrouter.com/en/main/route/error-element

can probably move a few things around here as well via router

src/app/router.tsx
```diff
@@ -50,6 +50,7 @@ import { EnvironmentActivityPage } from "@app/ui/pages/environment-detail-activi
 import { EnvironmentSettingsPage } from "@app/ui/pages/environment-detail-settings";
 import { EnvironmentIntegrationsPage } from "@app/ui/pages/environment-detail-integrations";
 import { EnvironmentBackupsPage } from "@app/ui/pages/environment-detail-backups";
+import { ErrorPage } from "@app/ui/pages/error";
```
neurosnap, line 53: import directly from `@app/ui` and it looks like we need to do the same for all the other imports above

neurosnap (approved):

PR #33, opened 2023-03-24
feat: [sc-10730] Fault tolerance for provisioning resources during FTUX

neurosnap:
Story details: https://app.shortcut.com/aptible/story/10730

### DIFF
--- a/src/deploy/database/index.ts
+++ b/src/deploy/database/index.ts
@@ -273,7 +273,6 @@ export const provisionDatabase = thunks.create<
       dbId: `${dbCtx.json.data.id}`,
       containerSize: 1024,
       diskSize: 10,
-      status: "queued",
       type: "provision",
     }),
   );
@@ -302,7 +301,6 @@ interface CreateDatabaseOpProps {
   containerSize: number;
   diskSize: number;
   type: "provision";
-  status: OperationStatus;
 }
 
 interface DeprovisionDatabaseOpProps {
@@ -322,12 +320,11 @@ export const createDatabaseOperation = api.post<
       }
 
       case "provision": {
-        const { containerSize, diskSize, type, status } = ctx.payload;
+        const { containerSize, diskSize, type } = ctx.payload;
         return {
           container_size: containerSize,
           disk_size: diskSize,
           type,
-          status,
         };
       }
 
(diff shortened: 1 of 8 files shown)


src/deploy/database/index.ts
```diff
@@ -273,7 +273,6 @@ export const provisionDatabase = thunks.create<
       dbId: `${dbCtx.json.data.id}`,
       containerSize: 1024,
       diskSize: 10,
-      status: "queued",
```
madhuravius, line 276: just sanity check, was this dropped because this is on the return?

src/deploy/operation/index.ts
```diff
@@ -37,6 +37,8 @@ export interface DeployOperationResponse {
   aborted: boolean;
   immediate: boolean;
   provisioned_iops: number;
+  disk_size: number;
+  container_size: number;
```
madhuravius, line 41: same as disk, if you choose to do it

src/deploy/database/index.ts
neurosnap: Yea we don't need these when submitting operations.

src/ui/pages/create-project-git.tsx
```diff
@@ -1444,13 +1469,36 @@ export const CreateProjectGitStatusPage = () => {
...
+            changes, and then push the code to us to redeploy. This screen will
+            track your deployment status.
+          </p>
+          <p className="text-black-500 my-4">
+            You can also try to trigger a redeploy now if you think it was a
+            transient error that caused the deployment to fail.
+          </p>
+
+          <Button
+            onClick={() => redeploy(true)}
```
madhuravius, line 1528: minor nit, is you might want to set a timeout on attempting to redeploy so people don't accidentally double click it, so set disabled and unset it after 5-10s

src/deploy/operation/index.ts
```diff
@@ -37,6 +37,8 @@ export interface DeployOperationResponse {
   aborted: boolean;
   immediate: boolean;
   provisioned_iops: number;
+  disk_size: number;
```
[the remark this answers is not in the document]
neurosnap: Let's save this for later.

src/ui/pages/create-project-git.tsx
neurosnap: Let's save this for testing.  The retry button will disappear almost immediately with its current function.

PR #40, opened 2023-03-29
fix: [sc-10271] Stack name empty after immediately creating account

neurosnap:
Story details: https://app.shortcut.com/aptible/story/10271

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -217,7 +217,11 @@ export const CreateProjectNamePage = () => {
     thunk.trigger();
   };
   const navigate = useNavigate();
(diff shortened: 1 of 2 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -217,7 +217,11 @@ export const CreateProjectNamePage = () => {
     thunk.trigger();
   };
   const navigate = useNavigate();
-  useQuery(fetchAllStacks());
+  const stacksQuery = useQuery(fetchAllStacks());
+
+  useEffect(() => {
```
madhuravius, line 222: 👍 interesting, i haven't been doing this and i need to look at a few things that probably could use this

PR #41, opened 2023-03-30
feat: activity/services/endpoints lists - sc-10735, sc-10736 sc-10180 (app/database/environment) (part 3 - focus on envs) 

madhuravius:
TODOs/questions to answer:

- [x] correct cli copy button (gives wrong command for logs)
- [x] change "Resources/Overview" to endpoints where applicable (dbs screen i think)
- [ ] rehydrate data on initial load (on all the layout screens?) - using hooks for this
- [ ] validate settings page (you can't really change env type from shared -> dedicated)
- [ ] ~should we drop integrations/certificates tab until we're ready?~ Adding services/integrations in read-only mode
- [x] we probably want to give the resource list views a little more love and match it closer to designs (being done in pt 4)

Resources List

image

Activity

We probably want to add some context for this with an optional parameter (so you can see which app/db it's targeting)

image

Backups List

We probably want to add some context for this with an optional parameter (so you can see which db/replica it's targeting)

image

Settings

image


src/deploy/endpoint/index.ts
```diff
@@ -117,6 +119,27 @@ export const selectEndpointsByAppId = createSelector(
   },
 );
 
+export const selectEndpointsByEnvironmentId = createSelector(
```
madhuravius, line 122: leave a todo here so we can create a backend ticket for this later

src/deploy/metric-drain/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    drainConfiguration: {},
+    createdAt: now,
+    updatedAt: now,
+    status: "pending",
+    ...md,
+  };
+};
+
+export const DEPLOY_METRIC_DRAIN_NAME = "metricDrains";
+const slice = createTable<DeployMetricDrain>({
```
neurosnap, line 54: I have some ideas on how to remove some of this boilerplate.  We should work on this eventually.

src/projects/index.ts
```diff
@@ -210,19 +210,6 @@ export const deployProject = thunks.create<CreateProjectSettingsProps>(
       env[`${db.env}_TMP`] = `{{${db.name}}}`;
     });
 
-    // Trigger configure operation now so we can store the env vars
```
neurosnap, line 213: Why was this deleted?

src/projects/index.ts
```diff
@@ -345,63 +329,27 @@ export const updateEnvWithDbUrls = thunks.create<{
...
     env: results.configure.json.data.env,
-    force,
   });
 
-  yield* put(setLoaderSuccess({ id }));
   yield next();
-  ctx.json = { message: "success" };
-});
-
-export const redeployApp = thunks.create<{
```
neurosnap, line 372: why was this deleted?

src/ui/pages/create-project-git.tsx
```diff
@@ -939,18 +939,6 @@ const Op = ({
     if (op.status === "failed") {
       return (
         <div className={cn(cns, "text-red")}>
-          {retry ? (
```
neurosnap, line 942: Why was this deleted?

src/ui/pages/create-project-git.tsx
```diff
@@ -991,66 +979,44 @@ const Op = ({
...
-}) => {
-  const dispatch = useDispatch();
+const AppStatus = ({ app }: { app: Pick<DeployApp, "id" | "handle"> }) => {
   const configOp = useSelector((s: AppState) =>
     selectLatestConfigureOp(s, { appId: app.id }),
   );
   const deployOp = useSelector((s: AppState) =>
     selectLatestDeployOp(s, { appId: app.id }),
   );
-  const retry = () => {
```
neurosnap, line 1039: Why was this deleted?

src/ui/shared/db/backup-list.tsx
```diff
@@ -0,0 +1,168 @@
...
+  IconEllipsis,
+  IconPlusCircle,
+  InputSearch,
+  LoadResources,
+  Loading,
+  ResourceListView,
+  TableHead,
+  Td,
+  tokens,
+} from "../";
```
neurosnap, line 20: Pull these from the actual module files instead of the root.  The reason is for creating a better module dependency graph.

neurosnap (changes requested):
It seems like a lot of unrelated code got deleted, any idea what's going on?

PR #42, opened 2023-03-30
Fix redeploy when no deploy op

neurosnap:
A couple of bug fixes I noticed when going through FTUX flow today.

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -1402,6 +1402,22 @@ export const CreateProjectGitStatusPage = () => {
     );
   };
 
+  // if the user started the deployment process but left before
+  // we could create an app deploy operation then we need to kick that
+  // off again here
+  useEffect(() => {
+    if (!hasDeployOperation(deployOp)) {
+      dispatch(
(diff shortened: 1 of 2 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -1402,6 +1402,22 @@ export const CreateProjectGitStatusPage = () => {
     );
   };
 
+  // if the user started the deployment process but left before
```
madhuravius, line 1405: clear enough to test and verify based on this comment if you're OOO tomorrow.  

my only kneejerk concern was this effectively firing on mount, but it may not have deployOp mounted (and mistakenly fires redeploy every time). I think what we would want is having a `deployOp.id` in the dependency array and/or ensuring that the deployOp query has already fired (if we know that to be true and there is no id, we must redeploy). this is likely a misinterpretation and i can probably rule it out when i get a min to test tomorrow

madhuravius: nice, lgtm or closer to it, still feeling weird but that comes with manual testing

madhuravius (approved):

PR #43, opened 2023-04-06
feat(ftux): allow user to choose stack when creating project

neurosnap:
This PR switches the FTUX flow to allow the user to select a stack to create the project under.  It also uses a shared `Select` component that we can tweak the design later down the road as-needed.

https://user-images.githubusercontent.com/1940365/230461722-02099249-0482-4ddb-89f4-ddf126b9ad4b.mov

Screenshot 2023-04-07 at 1 29 39 PM


src/ui/shared/form-group.tsx
```diff
@@ -28,36 +28,34 @@ export function FormGroupFeedback({ message, variant }: Feedback) {
 
 type Variant = "warn" | "danger" | "success" | "info";
 
-type Props = PropsWithChildren<{
+interface Props extends PropsWithChildren {
```
madhuravius, line 31: a question, not a request for change

threw me off and made me read: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

Looking at your change I can't see a strong reason why we changed this, just curious why (I think you can extend with the `&` bit anyway)

src/ui/shared/select-stacks.tsx
```diff
@@ -5,9 +5,9 @@ import { fetchAllStacks, selectStacksAsOptions } from "@app/deploy";
 
 import { EmptyResources, ErrorResources } from "./load-resources";
 import { Loading } from "./loading";
-import { SelectMenu } from "./select-menu";
+import { Select, SelectProps } from "./select";
 
-export const StackSelect = () => {
+export const StackSelect = (props: Omit<SelectProps, "options">) => {
```
madhuravius, line 10: TIL: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

neat

src/ui/shared/select.tsx
```diff
@@ -1,17 +1,32 @@
...
   defaultValue?: SelectOption;
   value?: SelectOption;
-  onSelect: (s: string) => void;
-};
+  onSelect: (s: SelectOption) => void;
+}
 
-export function Select({ value, options, onSelect, defaultValue }: Props) {
+export function Select({
+  value,
```
madhuravius, line 14: if there's only one choice we may want to try to disable it?

madhuravius:
Styling wise, if you can address it here it'd be probably be easier on the select menu, it looks like rounded corners, slightly larger down chevron and a description field?

image

src/ui/shared/select.tsx
neurosnap: I think you're right, I missed this in my code.  Good catch!

neurosnap: Actually, after looking at the code, I think the way we have it now is fine. The `select` dropdown will behave properly and it's work that's not really necessary.

madhuravius: understood

madhuravius (approved):

PR #45, opened 2023-04-07
design(ftux): remove prev and next buttons for production

neurosnap:
I also added progress text that we can continue to design

https://user-images.githubusercontent.com/1940365/230656339-e4761fdc-7062-4e03-85eb-d56a0bb5e4f6.mov

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -106,6 +106,7 @@ import {
   selectLatestScanOp,
   selectLatestSucceessScanOp,
 } from "@app/deploy/operation";
+import { selectEnv } from "@app/env";
 import { selectOrganizationSelected } from "@app/organizations";
 import {
   DbSelectorProps,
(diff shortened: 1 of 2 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -197,7 +216,7 @@ export const CreateProjectAddKeyPage = () => {
         </p>
       </div>
 
-      <FormNav next={createProjectAddNameUrl()} />
+      <FormNav cur={-1} next={createProjectAddNameUrl()} />
```
madhuravius, line 219: so -1 is disabled or hidden? -1 can feel really weird on frontend, why not another attribute that's just `disabled`? 

formnav seems fairly generic as a name, we should probably just have a disabledInProduction prop as well

madhuravius: did you plan on using a db column value (enumerable for this)?

neurosnap: `cur` is required anyway, `-1` is pretty common in the JS spec.

I can change the `FormNav` component name.

> did you plan on using a db column value (enumerable for this)?

Could you elaborate what you mean by this statement?

madhuravius: sure i'm okay with that, i just didn't see us use -1 anywhere else (usually accustomed to seeing disabled, etc for a component library) as it's more intuitive

-1 only really makes sense when using enumerables with a DB (as -1 it would USUALLY feel nonsensical to a person writing semantic code). `hidden`/`disabled`/etc boolean attributes are more common for me (or make more sense). i'm guessing here it makes more sense because we're talking about a paginated thing (or something that has a number representation), so it makes more sense here.

I'm okay with using -1 like this but we should probably shore it up and be consistent about it where pagination or numerical values make sense

madhuravius:
Rest of PR looks good, let me know if you need a reapproval after adding a `disabledInProduction` property or changing the component name

PR #47, opened 2023-04-07
enforce no unused locals and remove precommit hooks

neurosnap:
enables `noUnusedLocals` and removes precommit hooks


madhuravius (approved):
LGTM - i didn't do this as I assumed you were using the precommit hooks even though they were a tad annoying

PR #49, opened 2023-04-11
chore: ability to determine which app is running (ftux or nextgen)

neurosnap:
We already have an `origin` that we use to send to `auth-api` in order to determine which app the email verification links should redirect to so we can reuse that for determining if we are using the full-blown app or ftux.


madhuravius:
there is some overlap with my two branches on this one, we should probably just merge them up (https://github.com/aptible/app-ui/pull/51)

PR #59, opened 2023-04-20
fix(ftux): recover when db created but not provisioned

neurosnap:
fix(ftux): properly handle resource limit errors

Now when we create a database but fail to provision (e.g. resource limit error) we can properly recover from it by clearing resources and retrying.

### DIFF
--- a/src/deploy/database/index.ts
+++ b/src/deploy/database/index.ts
@@ -132,14 +132,20 @@ const { add: addDeployDatabases } = slice.actions;
 
 export const hasDeployDatabase = (a: DeployDatabase) => a.id !== "";
 export const databaseReducers = createReducerMap(slice);
-const initApp = defaultDeployDatabase();
-const must = mustSelectEntity(initApp);
+const initDb = defaultDeployDatabase();
+const must = mustSelectEntity(initDb);
 
 const selectors = slice.getSelectors(
   (s: AppState) => selectDeploy(s)[DEPLOY_DATABASE_NAME],
 );
 export const selectDatabaseById = must(selectors.selectById);
 export const { selectTableAsList: selectDatabasesAsList } = selectors;
+
+export const selectDatabaseByHandle = createSelector(
+  selectDatabasesAsList,
+  (_: AppState, p: { handle: string }) => p.handle,
+  (dbs, handle) => dbs.find((db) => db.handle === handle) || initDb,
+);
 
 export const selectDatabasesForTable = createSelector(
   selectDatabasesAsList,
@@ -251,9 +257,12 @@ export const createDatabase = api.post<
 });
 
 interface CreateDbResult {
-  dbCtx: Omit<DeployApiCtx<any, any>, "payload" | "json"> &
-    Payload<CreateDatabaseProps> &
-    FetchJson<DeployDatabaseResponse, any>;
+  dbCtx:
+    | null
+    | (Omit<DeployApiCtx<any, any>, "payload" | "json"> &
+        Payload<CreateDatabaseProps> &
+        FetchJson<DeployDatabaseResponse, any>);
+  dbId: string;
   opCtx: Omit<DeployApiCtx<any, any>, "payload" | "json"> &
     Payload<CreateDatabaseOpProps | DeprovisionDatabaseOpProps> &
     FetchJson<DeployOperationResponse, any>;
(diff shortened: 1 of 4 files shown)


src/deploy/database/index.ts
```diff
@@ -251,57 +257,72 @@ export const createDatabase = api.post<
 });
 
 interface CreateDbResult {
-  dbCtx: Omit<DeployApiCtx<any, any>, "payload" | "json"> &
-    Payload<CreateDatabaseProps> &
-    FetchJson<DeployDatabaseResponse, any>;
+  dbCtx:
+    | null
```
madhuravius, line 261: stylistic but there's nothing preceding this, how did this compile? 🤔

neurosnap: It's a typescript thing, they let you do it.

src/projects/index.ts
```diff
@@ -224,42 +224,47 @@ export const deployProject = thunks.create<CreateProjectSettingsProps>(
...
-              type: _type,
-              envId,
-              databaseImageId: db.imgId,
-            }),
-          );
-        })
-        .filter(Boolean),
+      dbs.map((db) => {
+        const handle = db.name.toLocaleLowerCase();
+        const _type = db.dbType;
```
madhuravius, line 229: nit - variable names with underscore usually means something else, but i don't mind (as in discardable value), migth wanna keep dbType

src/projects/index.ts
```diff
@@ -224,42 +224,47 @@ export const deployProject = thunks.create<CreateProjectSettingsProps>(
...
-      .map((res) => {
-        if (!res.json) return;
-        const { opCtx, dbCtx } = res.json;
-        if (!opCtx.json.ok) return;
-        if (!dbCtx.json.ok) return;
+    console.log(results);
+    const waiting = [];
+    for (let i = 0; i < results.length; i += 1) {
+      const res = results[i];
+      console.log("RES", res);
```
madhuravius, line 251: call out

src/projects/index.ts
```diff
@@ -224,42 +224,47 @@ export const deployProject = thunks.create<CreateProjectSettingsProps>(
...
      * now we hot-swap db env vars for the actual connection url
      */
 
-    const waiting = results
-      .map((res) => {
-        if (!res.json) return;
-        const { opCtx, dbCtx } = res.json;
-        if (!opCtx.json.ok) return;
-        if (!dbCtx.json.ok) return;
+    console.log(results);
```
madhuravius, line 247: ditto as below

src/projects/index.ts
neurosnap: fixed.

src/projects/index.ts
neurosnap: fixed.

src/projects/index.ts
neurosnap: fixed.

src/ui/pages/create-project-git.tsx
```diff
@@ -525,6 +529,13 @@ const starterTemplateOptions: StarterOption[] = [
     value: "git@github.com:aptible/template-laravel.git",
     query: { dbs: ["database_url:postgresql:14"] },
   },
+  {
+    label: "Deploy Demo App",
```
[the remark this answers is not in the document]
neurosnap: What's nice about this one is it is fully functional so I'd prefer to keep it around for the time being.

src/ui/pages/create-project-git.tsx
```diff
@@ -1068,6 +1082,31 @@ export const CreateProjectGitSettingsPage = () => {
   const [cmds, setCmds] = useState("");
   const cmdList = parseText(cmds, () => ({ id: "", http: false }));
 
+  // rehydrate already existing databases
+  // this allows us to trigger a provision operation if we failed to do so
+  useEffect(() => {
```
madhuravius, line 1087: if i read this correctly, it will reset, then begin rehydrating

but won't this just be on a loop forever? (since it'll reset, then it'll fetch, and the count will change again?)

src/ui/pages/create-project-git.tsx
```diff
@@ -1926,7 +1965,7 @@ const useProjectOps = ({ appId, envId }: { appId: string; envId: string }) => {
       [configOp, deployOp, ...provisionOps].filter((op) =>
         hasDeployOperation(op),
       ),
-    [configOp, deployOp, ...provisionOps],
+    [configOp, deployOp, provisionOps],
```
madhuravius, line 1968: 👍 this confused me on the last PR, but it's plural, so this looks like:

`[x, y, [z,z]]` or `[x, y, z]`, if it's the latter should probably be `provisionOp`

src/ui/pages/create-project-git.tsx
```diff
@@ -2089,10 +2132,11 @@ export const CreateProjectGitStatusPage = () => {
         )}
       </ProjectBox>
```
madhuravius, line 2134: this file is getting beefy, can't wait to break it up

src/ui/pages/create-project-git.tsx
```diff
@@ -1988,6 +2027,10 @@ export const CreateProjectGitStatusPage = () => {
   const redeployLoader = useSelector((s: AppState) =>
     selectLoaderById(s, { id: `${redeployApp}` }),
   );
+  const deployProjectLoader = useSelector((s: AppState) =>
+    selectLoaderById(s, { id: `${deployProject}` }),
```
madhuravius, line 2031: needless string interpolation?

src/deploy/database/index.ts
madhuravius: cool figured as much that it was stylistic, just threw me off

madhuravius:
there's a few borderline linter items in here we should probably add to our rules

src/ui/pages/create-project-git.tsx
neurosnap: `deployProject` is a function so we cast it to a string to get the endpoint name

src/ui/pages/create-project-git.tsx
neurosnap: It only resets the component state and then adds the existing databases.  It might get hit a few times but it won't refetch or loop weirdly.

madhuravius: TODO for us to add guards on this then maybe a limit of X calls before we don't let it retry without a hard reload or something

madhuravius (approved):
LGTM, didn't realize what

PR #61, opened 2023-04-24
fix(ftux): ux and bug fixes

neurosnap:
Mostly UX fixes.

- sc-12926

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -449,7 +449,7 @@ export const CreateProjectNamePage = () => {
           </FormGroup>
           <FormGroup
             label="Environment Name"
-            description="Lowercase alphanumerics, periods, and dashes only"
+            description="Lowercase alphanumerics, periods, dashes, and underscores only"
             htmlFor="name"
             feedbackVariant="info"
           >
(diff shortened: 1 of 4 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -1240,7 +1258,7 @@ export const CreateProjectGitSettingsPage = () => {
             htmlFor="envs"
             feedbackVariant={envErrors.length > 0 ? "danger" : "info"}
             feedbackMessage={envErrors.map((e) => e.message).join(". ")}
-            description="Environment Variables (each line is a separate variable in format: ENV_VAR=VALUE)."
+            description="Add any additional required variables, such as API keys, KNOWN_HOSTS setting, etc. Each line is a separate variable in format: ENV_VAR=VALUE."
```
madhuravius, line 1261: if descriptions support `pre` type elements, we should use that for the caps here

madhuravius: like `<span> .... <pre>SOMETHING=true</pre></span>` instead of plaintext

neurosnap: @joeclayallday asked me to remove those at one point.

madhuravius (approved):
nitpick on better descriptions

lgtm otherwise, mostly copy edits it seems

PR #62, opened 2023-04-24
chore(ftux): mark onboarding_status based on start instead of success

neurosnap:


madhuravius (commented):
do we want to add backend statuses or changes as a result of this?

neurosnap:
I honestly don't think it's necessary, but I'll defer to your judgement

madhuravius:
the only reason i was asking was because this won't work unless it fits certain statuses.

if this works fine against backend we're g2g, i think a simple pass with you is enough for approval

neurosnap:
#63 supersedes

PR #63, opened 2023-04-25
onboarding status, service command fixes, and ux fixes

neurosnap:
https://app.shortcut.com/aptible/story/12919/onboarding-status-should-denote-start-of-operation-not-success


src/deploy/app/index.ts
```diff
@@ -277,28 +277,6 @@ export const createAppOperation = api.post<AppOpProps, DeployOperationResponse>(
     const body = getBody();
     ctx.request = ctx.req({ body: JSON.stringify(body) });
     yield next();
-
```
madhuravius, line 280: if we remove this how will the page behave idempotently (or attempt to)?

madhuravius (approved):

PR #65, opened 2023-04-26
chore: production app.aptible.com and sbx-main setup

neurosnap:


madhuravius (commented):
aren't there other places we gotta update too? .env.example, ARGs, and readme?

or is the desire to mostly point to nextgen for developers and app for customer facing?

Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN env | grep VITE
 
 RUN yarn build
 
-FROM nginx:1.13.11 as nginx
+FROM nginx:1.24.0 as nginx
```
madhuravius, line 25: 👍

madhuravius (approved):

PR #70, opened 2023-04-27
fix(ftux): origin can only be `nextgen` or `app`

neurosnap:
The value of `origin` is important to `auth-api` so we want it to be the "final" value so we don't have to change it after the initial implementation.

Related: https://github.com/aptible/auth-api/pull/500

### DIFF
--- a/src/types/state.ts
+++ b/src/types/state.ts
@@ -25,7 +25,7 @@ export interface Env {
   apiUrl: string;
   sentryDsn: string;
   legacyDashboardUrl: string;
-  origin: "nextgen" | "ftux";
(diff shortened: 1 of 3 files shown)


madhuravius (approved):
LGTM - i'm okay with this, i'm guessing this is a consequence of this PR too: https://github.com/aptible/app-ui/pull/65/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R7

sorry you gotta go back and forth on this, aside from grep i can't easily catch these

edit - also may want to two-way link that other auth-api pr so people have a link to this one later

PR #71, opened 2023-04-27
fix(ftux): filter stack selector by org if stack has org

neurosnap:


src/deploy/stack/index.ts
```diff
@@ -86,6 +86,21 @@ export const selectStacksAsOptions = createSelector(
...
+  selectStacksAsList,
+  (_: AppState, p: { orgId: string }) => p.orgId,
+  (stacks, orgId) => {
+    return stacks
+      .filter(
+        (stack) =>
+          stack.organizationId === "" || stack.organizationId === orgId,
+      )
+      .map(stackToOption)
+      .sort((a, b) => a.label.localeCompare(b.label));
```
madhuravius, line 100: you sure it's `label`? i swore it was `handle` or actually in this case unfortunately it's `name`

madhuravius: i'm guessing it's the `option` that changes this, nvm

madhuravius (approved):

PR #76, opened 2023-04-27
fix(ftux): only select single provision op for each resource

neurosnap:


src/deploy/operation/index.ts
```diff
@@ -169,10 +169,20 @@ export const selectOperationsByResourceId = createSelector(
   (ops, resourceId) => ops.filter((op) => op.resourceId === resourceId),
 );
 
-export const selectOperationsByResourceIds = createSelector(
+export const selectLatestOperationByResourceIds = createSelector(
```
madhuravius, line 172: correct me if i'm wrong but title is wrong to commit

madhuravius (approved):
LGTM - but `resourceIds` should always be an array type, right? if it's ever null could cause a bad thing to happen

PR #78, opened 2023-04-27
fix: dont send local storage token to /current_token

neurosnap:
# What?

We were always sending `localStorage.token` to `/current_token` and `auth-api` would return that token ... even if the `session_token` had a more recent one.

This ensures we never send `/current_token` the token stored in local storage.

- [current_token_id](https://github.com/aptible/auth-api/blob/c26e460bd13d7f1d9ab8f814d50ee4d772305512/app/controllers/application_controller.rb#L10)
- [current_token](https://github.com/aptible/fridge/blob/38fadfb6b515fdf29d8e537412a8a32424bedddd/lib/fridge/rails_helpers.rb#L22)
- [session_token](https://github.com/aptible/fridge/blob/38fadfb6b515fdf29d8e537412a8a32424bedddd/lib/fridge/rails_helpers.rb#L43)

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -34,6 +34,7 @@ export type DeployApiCtx<P = any, S = any> = ApiCtx<P, S, { message: string }>;
 export interface AuthApiCtx<P = any, S = any>
   extends ApiCtx<P, S, AuthApiError> {
   elevated: boolean;
+  noToken: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


madhuravius:
this sort of makes sense, but does that mean we do not invalidate some types of tokens?

madhuravius (approved):

PR #79, opened 2023-04-28
fix(login): always fetch

neurosnap:
We weren't fetching all the on load data after the user logs in.


madhuravius (approved):
LGTM - we may want to worry about debouncing these calls as I'm scared of this one given the possible volume of calls to auth

PR #80, opened 2023-04-28
fix(ftux): recover from db failing to provision

neurosnap:
Repro:
- Try to provision a DB
- Plan limits exceeded
- Deprovision resources
- Go back
- Try to save and deploy again
- No db operation in the status list

Other Fixes:
- sort orgs by updated_at
- properly split env var text area

### DIFF
--- a/src/organizations/index.ts
+++ b/src/organizations/index.ts
@@ -102,18 +102,22 @@ function deserializeOrganization(o: OrganizationResponse): Organization {
 }
 
 type FetchOrgCtx = AuthApiCtx<
+  {},
   HalEmbedded<{ organizations: OrganizationResponse[] }>
 >;
-export const fetchOrganizations = authApi.get(
+export const fetchOrganizations = authApi.get<FetchOrgCtx>(
   "/organizations",
   { saga: cacheTimer() },
-  function* onFetchOrgs(ctx: FetchOrgCtx, next) {
+  function* onFetchOrgs(ctx, next) {
     yield next();
     if (!ctx.json.ok) {
       return;
     }
 
-    const orgs = ctx.json.data._embedded.organizations;
+    const orgs = ctx.json.data._embedded.organizations.sort(
+      (a, b) =>
+        new Date(b.updated_at).getTime() - new Date(a.updated_at).getTime(),
+    );
     if (orgs.length > 0) {
       ctx.actions.push(setOrganizationSelected(orgs[0].id));
     }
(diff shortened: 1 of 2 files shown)


src/organizations/index.ts
```diff
@@ -102,18 +102,22 @@ function deserializeOrganization(o: OrganizationResponse): Organization {
...
+  function* onFetchOrgs(ctx, next) {
     yield next();
     if (!ctx.json.ok) {
       return;
     }
 
-    const orgs = ctx.json.data._embedded.organizations;
+    const orgs = ctx.json.data._embedded.organizations.sort(
+      (a, b) =>
+        new Date(b.updated_at).getTime() - new Date(a.updated_at).getTime(),
```
madhuravius, line 119: should prob do by another value like a name maybe

madhuravius: or the id itself, one can't mutate the id, but org might get updated after creation

src/ui/pages/create-project-git.tsx
```diff
@@ -1164,11 +1164,13 @@ export const CreateProjectGitSettingsPage = () => {
         const img = dbImages.find((i) => i.id === db.databaseImageId);
         if (!img) return;
         const env = envList.find((e) => e.value === `{{${db.handle}}}`);
-        if (!env) return;
+        if (!env) {
```
madhuravius, line 1167: wonder if this is how we killed our express apps

src/organizations/index.ts
neurosnap: I wanted the most recently used org to be the first selected.  I'm not able to switch to the prod Aptible org because some other one is ahead in the list.

madhuravius: we should just leave a note about this so we don't get stuck into this (as multi-org isn't something we considered much on this new ui)

madhuravius (dismissed):

madhuravius:
> Try to provision a DB
> Plan limits exceeded
> Deprovision resources
> Go back
> Try to save and deploy again
> No db operation in the status list

If you never leave the page it still works out

src/ui/pages/create-project-git.tsx
```diff
@@ -727,10 +727,11 @@ const parseText = <
     .split("\n")
     .map(trim)
     .map((t) => {
-      const vals = t.split("=").map(trim);
+      const [key, ...values] = t.split("=").map(trim);
```
madhuravius, line 730: apparently it was this (not on this PR), that prompted below change:

```
that will hydrate every time
on hitting backbutton
how do we avoid that
```

madhuravius: this is done becuase values might be X="Y=Z" as some environment variables may have equal signs

madhuravius (approved):

PR #81, opened 2023-04-28
chore: saga-query logging

neurosnap:


src/api/index.ts
```diff
@@ -150,6 +160,7 @@ const MINUTES = 60 * SECONDS;
 export const cacheTimer = () => timer(5 * MINUTES);
 
 export const api = createApi<DeployApiCtx>();
+api.use(debugMdw);
```
madhuravius, line 163: shouldn't we only use this if config is enabled for debug? or add that in line 48 as a guard to prevent this showing up?

neurosnap: `debug` is a conditional logger.  I guess we could check for dev but i also think it's fine to leave it as-is since opening up the console will not spit these out

madhuravius: low enough impact doesn't matter yet

madhuravius (dismissed):
LGTM - instructions already exist: `localStorage.debug = 'app:*';`

madhuravius (dismissed):

madhuravius (approved):

PR #87, opened 2023-05-02
feat(ftux): resume app flow

neurosnap:
This allows us to direct users through the app creation flow without needing to create an environment.  If the environment already has databases then we won't let them create any new ones.


src/ui/pages/create-project-git.tsx
```diff
@@ -240,6 +240,49 @@ export const CreateProjectSetupPage = () => {
...
+    if (hasDeployOperation(deployOp)) {
+      navigate(createProjectGitStatusUrl(appId));
+      return;
+    } else if (hasDeployOperation(scanOp) && scanOp.status === "succeeded") {
+      navigate(createProjectGitSettingsUrl(appId));
+      return;
+    }
+
+    navigate(createProjectGitPushUrl(appId));
+  }, [env.id, app.id, appOps]);
```
madhuravius, line 281: probalby want deployOp in dependency array or its id

madhuravius (approved):

madhuravius (approved):

PR #89, opened 2023-05-03
Webauthn

neurosnap:
This PR does two things:
- Fixes a bug with logging in via webauthn
- Ability to register security keys using app-ui

### DIFF
--- a/src/ui/pages/login.tsx
+++ b/src/ui/pages/login.tsx
@@ -88,7 +88,7 @@ export const LoginPage = () => {
       dispatch(
         loginWebauthn({
           ...loginPayload,
-          webauthn: loader.meta.exception_context.u2f,
+          webauthn: loader.meta.exception_context.u2f?.payload,
         }),
       );
     }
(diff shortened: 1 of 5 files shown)


src/ui/pages/login.tsx
```diff
@@ -88,7 +88,7 @@ export const LoginPage = () => {
       dispatch(
         loginWebauthn({
           ...loginPayload,
-          webauthn: loader.meta.exception_context.u2f,
+          webauthn: loader.meta.exception_context.u2f?.payload,
```
neurosnap, line 91: This was the bug

madhuravius: 😠 wow

is there an easy way to test this btw along with some of the other flows when we're closer to ready?

madhuravius: also i can't see the types on this pr for `loginWebauthn`, shouldn't that ideally have saved us?

neurosnap: The types are created by us so we if we don't type things properly then you can get a runtime error.

Something like `zod` probably could have saved us here.

madhuravius: ✅  - tested on sbx main
looks like an error on typing on OUR end pre-send to auth api

madhuravius (approved):

PR #90, opened 2023-05-03
sc-10735-sc-10736-sc-10180 - collected read only changes

neurosnap:


madhuravius:
A few things are badly styled now (blue links for example), but not intrusive (current setup works fine or so it seems):

image

Was able to go through the core flow with minimal issue. The "completed pages" which I spent about 2/3 of a sprint on are pretty awkward, in that they don't really look as nice. Similarly there are a few awkward behaviors with those pages (componentDidMount behavior needs to be corrected and they dont' fully fetch all the data they need yet)

madhuravius:
I'm going to merge this, and put up another for the missing bits i had local (log drains + metric drains), and we can come back to these pages afterwards

madhuravius (approved):

PR #91, opened 2023-05-06
ftux deployments by app

neurosnap:
Fixes:
- Deployments page now has same status pill as status page
- Deployments page now has correct "finish setup" link
- Deployments are listed by `App` instead of `Account`
- If there are 0 operations for an `Account` we resolve that status to `unknown` instead of `succeeded`

# BEFORE
Screenshot 2023-05-06 at 3 24 51 PM

# AFTER
Screenshot 2023-05-06 at 3 24 55 PM

### DIFF
--- a/src/deploy/app/index.ts
+++ b/src/deploy/app/index.ts
@@ -15,7 +15,11 @@ import type {
 import { createAction, createSelector } from "@reduxjs/toolkit";
 import { call, createThrottle, poll, select } from "saga-query";
 
-import { findEnvById, selectEnvironments } from "../environment";
+import {
+  findEnvById,
+  hasDeployEnvironment,
+  selectEnvironments,
+} from "../environment";
 import { deserializeImage } from "../image";
 import { deserializeDeployOperation, waitForOperation } from "../operation";
 import { selectDeploy } from "../slice";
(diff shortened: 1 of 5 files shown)


src/deploy/operation/index.ts
```diff
@@ -273,6 +273,11 @@ export const fetchEnvOperations = api.get<EnvOpProps>(
   "/accounts/:envId/operations?page=:page",
 );
 
+export const fetchAllEnvOps = thunks.create<EnvIdProps>(
+  "fetch-all-env-ops",
+  combinePages(fetchEnvOperations, { max: 5 }),
```
madhuravius, line 278: we might want to keep fetching if the most recent ops are backups, that thing eats up a lot of pages even after just a week or two i think

madhuravius (dismissed):

src/deploy/app/index.ts
```diff
@@ -166,6 +170,24 @@ export const selectAppsForTableSearch = createSelector(
...
+  selectEnvironments,
+  selectAppsAsList,
+  (envs, apps) => {
+    return apps.filter((app) => {
+      const env = findEnvById(envs, { id: app.environmentId });
+      if (!hasDeployEnvironment(env)) {
+        return false;
+      }
+      if (env.onboardingStatus === "unknown") {
+        return false;
```
madhuravius, line 183: we might want to leave a note somewhere that we only will show environments deployed with our new flow (aka legacy/v1 is not displayed on this list)

src/deploy/operation/index.ts
neurosnap: Yeah I'm just worried about over-fetching.  I can increase it to 10 pages?

src/deploy/app/index.ts
neurosnap: Good idea

src/deploy/operation/index.ts
madhuravius: we can leave it at 5 but maybe make a ticket for us to have a paginator that just keeps going and collects until X non backup operations found?

madhuravius: or 5 instantly and come back somehow in a throttled way AFTER 5 until it finds something?

madhuravius (approved):

PR #94, opened 2023-05-09
chore: added tuna tracking to app-ui to avoid possible blind spots in tracking users across dashboard

madhuravius:
Other things we probably want to do:

* ~probably need to review w/ security team on this stuff because we do use cookies and we probably need to disclose it (esp. with tuna)~
* ~for now only enabling pageview event, but we may want to return to other portions~
* ~we could probably react-ify some parts of tuna, not sure if we want to~ (later)

---

Flag enabled:

image

Flag disabled:

image

---

Cookie notice aesthetic (disabled for now):

image


src/app/app.tsx
```diff
@@ -6,15 +6,19 @@ import type { AppState } from "@app/types";
 import { ModalPortal, StandaloneErrorBoundary } from "@app/ui";
 
 import { ftuxRouter, router } from "./router";
+import { Tuna } from "./tuna";
 import { selectOrigin } from "@app/env";
 import { RouterProvider } from "react-router";
+// import { CookieNotice } from "@app/ui/shared/cookie-notice";
```
madhuravius, line 12: can remove this but need to check item on checklist before this can be done

src/app/tuna.tsx
```diff
@@ -0,0 +1,19 @@
+"use client";
+
+import { useEffect } from "react";
+
+export const Tuna = () => {
+  useEffect(() => {
+    if (process.env.NODE_ENV !== "production") {
```
madhuravius, line 7: this var surprisingly works, not sure how, may be reserved when running on local

src/ui/shared/cookie-notice.tsx
```diff
@@ -0,0 +1,129 @@
...
+      return;
+    }
+
+    if (getCookie(COOKIE_NAME) === undefined) {
+      setShowNotice(true);
+    }
+  }, []);
+
+  const closeFn = () => {
+    writeCookie(COOKIE_NAME, "1", 90);
```
madhuravius, line 90: this doesn't actually seem to fully work, need to revisit

index.html
```diff
@@ -9,5 +9,14 @@
   <body>
     <div id="app"></div>
     <script type="module" src="/src/app/init.ts"></script>
+    <script id="tuna">
+      (function(a,p,t,i,b,l,e) {
```
madhuravius, line 13: instead of modifying this directly, i've set this via templates (if you want a cleaner index.html). it's kinda annoying and honestly probably not worth the time to do

src/app/tuna.tsx
neurosnap: You want `import.meta.env.MODE` or just `import.meta.env.DEV` I believe: https://vitejs.dev/guide/env-and-mode.html#env-variables

src/ui/shared/cookie-notice.tsx
neurosnap: Why do we need this?

index.html
```diff
@@ -9,5 +9,14 @@
   <body>
     <div id="app"></div>
     <script type="module" src="/src/app/init.ts"></script>
+    <script id="tuna">
```
neurosnap, line 12: I wonder if we can exclude this from our build if tuna is disabled as well?

src/ui/shared/cookie-notice.tsx
madhuravius: i think if we wanted to show a cookie notice (which [we might be obligated to do](https://aptible.slack.com/archives/C0237PYRF6D/p1683637574878789)), we have to know to close it and keep it closed (or it will keep appearing on every page)

src/app/tuna.tsx
madhuravius: ill do that instead, process.env worked and we used it for something else

index.html
madhuravius: not really worth the squeeze imo, doesn't fire otherwise, we'd have to do conditional templating, which has never been pleasant in my experience

src/app/tuna.tsx
madhuravius: used to check redux inspector change as well:

dev

image

prod 

image

index.html
madhuravius: if you feel strongly, we can use something like this: https://stackoverflow.com/questions/70818545/how-to-include-html-partials-using-vite

src/ui/shared/cookie-notice.tsx
```diff
@@ -0,0 +1,129 @@
...
+      continue;
+    }
+
+    cookies[tokens[0].trim()] = tokens[1].trim();
+  }
+
+  return cookies;
+};
+
+const allParams = (): Params => {
```
madhuravius, line 52: see if these params funcs can be removed

madhuravius:
I verified this works locally against `.aptible.com` domains (dismiss cookie) and removed all unnecessary parts (copied from another repo)

index.html
madhuravius: we're not doing this now

neurosnap (approved):

PR #99, opened 2023-05-10
feat: added freeform text inputs and choice survey field forms, sc-13222, sc-13223

madhuravius:
- [ ] Group requests and promise-all equivalent them with saga-query (for multi select survey)
- [ ] Survey freeform submission against tuna event

In both cases, todos remaining in code to actually post to backend

---

Free form text input

image

Multi select survey form

image

### DIFF
--- a/src/ui/pages/create-project.tsx
+++ b/src/ui/pages/create-project.tsx
@@ -1,11 +1,15 @@
 import {
   AptibleLogo,
   Box,
+  Button,
+  ButtonIcon,
   ButtonLink,
   ExternalLink,
   IconArrowRight,
+  IconCheck,
   IconChevronDown,
   IconChevronUp,
+  IconThumbsUp,
   tokens,
 } from "../shared";
 import { createProjectGitUrl } from "@app/routes";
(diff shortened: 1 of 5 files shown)


src/feedback/index.ts
```diff
@@ -0,0 +1,23 @@
+import { createAssign, createReducerMap } from "@app/slice-helpers";
+import { AppState, Feedback } from "@app/types";
+
+export const createFeedback = (e: Partial<Feedback> = {}): Feedback => {
+  return {
```
neurosnap, line 5:
```ts
{
    preDeploySurveyAnswered: false,
    freeformFeedbackGiven: false,
    ...e,
};
```

src/ui/pages/create-project.tsx
```diff
@@ -126,6 +136,125 @@ const CreateProjectFooter = () => {
...
+  const preDeploySurveyAnswered = useSelector(selectPreDeploySurveyAnswered);
+  const [surveyDockerPushSet, setSurveyDockerPushSet] =
+    useState<boolean>(false);
+  const [surveyDockerComposeSet, setSurveyDockerComposeSet] =
+    useState<boolean>(false);
+  const [surveyGithubIntegrationSet, setSurveyGithubIntegration] =
+    useState<boolean>(false);
+  const [feedbackSubmitted, setFeedbackSubmitted] = useState<boolean>(false);
+
+  const viewSurveyForm = () => {
```
neurosnap, line 150: This might be better as a separate component.  Thoughts?

PR #105, opened 2023-05-11
feat(ftux): show current vhosts on status page

neurosnap:
This should help users better understand what they are provisioning as well as give them some information to debug if they selected the incorrect service to provision and HTTPS endpoint.
Screenshot 2023-05-10 at 11 42 12 PM

### DIFF
--- a/src/app/test/create-project.test.tsx
+++ b/src/app/test/create-project.test.tsx
@@ -78,7 +78,7 @@ describe("Create project flow", () => {
       expect(ops.length).toEqual(3);
 
       // create https endpoint
-      await screen.findByText("Which service needs an endpoint?");
+      await screen.findByText(/Which service needs an/);
 
       const vhostSelector = await screen.findAllByRole("radio");
       fireEvent.click(vhostSelector[0]);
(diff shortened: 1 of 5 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -2341,12 +2349,25 @@ export const CreateProjectGitStatusPage = () => {
...
+          <ExternalLink
+            href={`${legacyUrl}/apps/${app.id}/vhosts`}
+            variant="info"
+          >
+            Manage Endpoints
+          </ExternalLink>
+        </StatusBox>
+      ) : (
         <StatusBox>
           <h4 className={tokens.type.h4}>Which service needs an endpoint?</h4>
```
madhuravius, line 2367: we could probably include a link to explaining endpoints here so people aren't confused potentially or know which services to apply it to

madhuravius (dismissed):

src/app/test/create-project.test.tsx
```diff
@@ -19,7 +19,7 @@ describe("Create project flow", () => {
       render(<App />);
 
       // deploy code landing page
-      const el = await screen.getByRole("button", {
+      const el = screen.getByRole("button", {
```
madhuravius, line 22: interesting, this should be awaited i thought

neurosnap: linter complained

PR #110, opened 2023-05-17
chore: update dependencies

neurosnap:


madhuravius (approved):
LGTM - some typing removals that are probably falling back to implicit defaults?

madhuravius:
also i'm sure you tested but the type=submit changes can have adverse outcomes, you should verify intention there and/or e.preventDefault if necessary on those forms

neurosnap:
Yep, confirmed they work as expected

PR #111, opened 2023-05-17
chore: added timer on operations to prevent user from thinking the page is frozen (sc-13947)

madhuravius:
While not completed or failed:

image - updates every 1s

When completed/failed:

image

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -18,7 +18,7 @@ import {
   useQuery,
 } from "saga-query/react";
 
-import { prettyDateRelative } from "@app/date";
+import { prettyDateRelative, timeBetween } from "@app/date";
 import {
   appDetailUrl,
   createProjectAddKeyUrl,
@@ -142,7 +142,6 @@ import {
   deployProject,
   redeployApp,
 } from "@app/projects";
-import { timeBetween } from "@app/date";
 
 export const CreateProjectLayout = ({
   children,
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -142,6 +142,7 @@ import {
   deployProject,
   redeployApp,
 } from "@app/projects";
+import { timeBetween } from "@app/date";
 
 export const CreateProjectLayout = ({
   children,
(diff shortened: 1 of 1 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -1472,7 +1472,25 @@ const Op = ({
   alwaysRetry?: boolean;
   status: OperationStatus;
 }) => {
+  const [runningTime, setRunningTime] = useState<string | null>(null);
   const [isOpen, setOpen] = useState(false);
+
+  useEffect(() => {
+    if (!["succeeded", "failed"].includes(op.status)) {
```
neurosnap, line 1479: `if (["succeeded", "failed"].includes(op.status)) return;`

src/ui/pages/create-project-git.tsx
```diff
@@ -1503,6 +1525,11 @@ const Op = ({
               Re-run
             </Button>
           ) : null}
+          {completedTime && (
```
neurosnap, line 1528: could we do a ternary instead here?  The `&&` is awkward

src/ui/pages/create-project-git.tsx
```diff
@@ -1472,7 +1472,25 @@ const Op = ({
...
   status: OperationStatus;
 }) => {
+  const [runningTime, setRunningTime] = useState<string | null>(null);
   const [isOpen, setOpen] = useState(false);
+
+  useEffect(() => {
+    if (!["succeeded", "failed"].includes(op.status)) {
+      const interval = setInterval(() => {
+        setRunningTime(
+          op.createdAt
```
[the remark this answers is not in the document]
madhuravius: not createdat, but updated at and was being overly cautious, can dump it

src/ui/pages/create-project-git.tsx
madhuravius: sure, can do this

madhuravius: we should probably add this to lint rules imo somehow as this can be autorefactored (early returns)

src/ui/pages/create-project-git.tsx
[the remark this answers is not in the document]
madhuravius: interesting 

```
export interface Timestamps {
  createdAt: string;
  updatedAt: string;
}
```

is never optional either

src/ui/pages/create-project-git.tsx
madhuravius: dropped this entirely.

i think i was running into issues on local state when i had a bad connection, but this is always set

neurosnap (approved):

PR #112, opened 2023-05-17
feat: toasts

neurosnap:
https://github.com/aptible/app-ui/assets/1940365/f42ef0cf-21f7-4428-9efe-a8e16ee6df4b

### DIFF
--- a/src/types/toast.ts
+++ b/src/types/toast.ts
@@ -1,0 +1,12 @@
+import { StatusVariant } from "./helpers";
+
+export interface Toast {
+  id: number;
+  type: StatusVariant;
+  text: string;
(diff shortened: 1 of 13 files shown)


src/types/toast.ts
```diff
@@ -0,0 +1,12 @@
+import { StatusVariant } from "./helpers";
+
+export interface Toast {
+  id: number;
+  type: StatusVariant;
+  text: string;
+  cta?: () => void;
```
madhuravius, line 7: we should prob avoid acronyms if we can, harder to search etc

neurosnap:
Not implementing this just yet, but can use it as a reference point later.

PR #115, opened 2023-05-18
style(ftux): settings page

neurosnap:
Screenshot 2023-05-18 at 1 57 08 PM

### DIFF
--- a/src/test/index.tsx
+++ b/src/test/index.tsx
@@ -14,10 +14,13 @@ import { ftuxRoutes } from "@app/app/router";
 import { bootup } from "@app/bootup";
 import { testEnv } from "@app/mocks";
 import type { AppState } from "@app/types";
-import { configureStore } from "@reduxjs/toolkit";
+import { Store, configureStore } from "@reduxjs/toolkit";
+import { waitFor } from "@testing-library/react";
 import { REHYDRATE } from "redux-persist";
 
-export const setupTestStore = (initState: Partial<AppState> = {}) => {
+export const setupTestStore = (
+  initState: Partial<AppState> = {},
+): { store: Store<AppState> } => {
   const middleware = [];
   const prepared = prepareStore({
     reducers: reducers,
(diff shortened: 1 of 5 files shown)


src/test/index.tsx
```diff
@@ -35,7 +38,7 @@ export const setupTestStore = (initState: Partial<AppState> = {}) => {
 
   prepared.run();
 
-  return { store };
+  return { store: store as any };
```
madhuravius, line 41: why not just have more permissive typing on unit tests? doesn't really do much if we do stuff like this here, right? i think we can omit it for testing by glob

src/ui/pages/create-project-git.tsx
```diff
@@ -1313,32 +1316,36 @@ export const CreateProjectGitSettingsPage = () => {
...
+                  {existingDbs.length > 0 ? (
+                    <p>
+                      Databases have already been created so you cannot make
+                      changes to them in this screen anymore.
+                    </p>
+                  ) : null}
+                </div>
+              }
+            >
+              {existingDbs.length > 0 ? null : (
```
madhuravius, line 1338: will existingDbs have a length if loading?

why not just have an early return if loading if that's NOT the case? you suggested in another PR to me:

```ts
if (imgLoader.isInitialLoading) return ()

...

existing formgroup here
```

neurosnap: We don't want to block the entire Page from rendering for missing one piece of data.  We also won't need this data if the user already has databases, since they won't be allowed to pick them.

madhuravius: i'd probably pull this into a separate render function then, it has a few nested ternaries that reminds me of your advice about existential checks (which is part of my confusion here, is it possible for something to NOT be loading, but still have existing dbs?)

either way at this point it's more of a nit based on your changes

madhuravius (approved):
LGTM if reappr needed

src/ui/pages/create-project-git.tsx
neurosnap: Good idea!  I'm going to add a card to our sprint to break this file apart.  At that point we can figure out a good way to break things up.  Right now that component uses a lot of dependencies so I want to make sure we spend the right amount of time breaking things up properly.

PR #123, opened 2023-05-23
feat: impersonate a user

neurosnap:
Screenshot 2023-05-24 at 3 02 45 PM


src/ui/shared/banner.tsx
```diff
@@ -68,7 +68,7 @@ export const BannerMessages = ({
   if (isSuccess) {
     return (
       <Banner className={className} variant="success">
-        {message}
+        {message || "Success!"}
```
madhuravius, line 71: this feels weird, can we pipe this up from the component and not show a banner if no message at all? feels wrong

src/ui/pages/impersonate.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  if (sso) {
+    return {
+      subjectToken: email,
+      subjectTokenType: "aptible:user:email",
+      ssoOrganization: orgHref,
+    };
+  }
+
+  if (email) {
```
madhuravius, line 34: in other places you're pretty explicit about `string === ""` checks we should prob be consistent

src/reset-store/index.ts
```diff
@@ -28,12 +28,22 @@ const keepState = (
 export const resetReducer =
   (rootReducer: Reducer, persistConfig: PersistConfig<any>) =>
   (state: AppState | undefined, action: Action<any>) => {
-    if (action.type === `${resetStore}`) {
+    let reset = false;
+    if (action.type === BATCH) {
+      const actions = (action as any).payload as Action[];
+      actions.forEach((act) => {
+        if (act.type === `${resetStore}`) {
+          reset = true;
```
madhuravius, line 36: nitpick but you can probably do an early break with a for loop here instead as you only care about the first encounter i think

src/app/router.tsx
```diff
@@ -409,6 +410,10 @@ export const appRoutes: RouteObject[] = [
             path: routes.TEAM_PATH,
             element: <TeamPage />,
           },
+          {
```
madhuravius, line 413: not sure if you want to do it in a separate PR but i think it's an easy win to put a impersonate banner on the root component of a layout or shared parent (or something where we inject it to the top of the dom of each component) that shows that someone is impersonated as in this [slack thread](https://aptible.slack.com/archives/C04PTPE7ZB6/p1684957188454429?thread_ts=1684954976.463099&cid=C04PTPE7ZB6) (low LOE and people will probably like it)

madhuravius:
Sanity check:

* When an impersonation token expires, do we also kick a user out (?) - i hope so (and it doesn't fall back to parent)
* When an impersonation is active and we switch between uis (we'll have two way linking for a while anyway), I want to make sure that settings pages/others are very clearly not being messed up, i can see a case where support is helping someone and if we go back to the wrong account they might apply changes to the incorrect account

src/app/router.tsx
madhuravius: this is separate pr

madhuravius (approved):

PR #126, opened 2023-05-31
Activity Page

neurosnap:


src/deploy/activity/index.ts
```diff
@@ -0,0 +1,109 @@
...
+        return { ...op, envHandle: env.handle, resourceHandle };
+      })
+      .sort(
+        (a, b) =>
+          new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime(),
+      ),
+);
+
+const MAX_RESULTS = 50;
+// TODO: remove `slice`
```
madhuravius, line 42: reminder

neurosnap: Yeah that'll be for later.  I want to keep it to 50 for now.  I can remove the comment I guess?

madhuravius: for sure i just realized as i saw it in other files so you can ignore my comment here

src/deploy/environment/index.ts
```diff
@@ -138,6 +138,18 @@ export const {
...
+// TODO: remove `slice`
+export const selectEnvsAsIds = createSelector(
+  selectEnvironmentsAsList,
+  (envs) =>
+    envs
+      .sort(
+        (a, b) =>
+          new Date(b.updatedAt).getTime() - new Date(a.updatedAt).getTime(),
+      )
+      .slice(0, Math.min(envs.length, 5))
```
madhuravius, line 150: maybe i'm misreading this but isn't this only taking a maximum of 0 to 5 items sorted by udpated?

the title does not reflect that, maybe we should either supply that as a param or as a variable? (or reflect in the name)

src/ui/shared/log-viewer.tsx
```diff
@@ -0,0 +1,81 @@
+import { useEffect } from "react";
+import { useSelector } from "react-redux";
+import { selectDataById } from "saga-query";
+import { useApi } from "saga-query/react";
+
+import { fetchOperationLogs } from "@app/deploy";
+import { AppState, DeployOperation } from "@app/types";
+
+const LogLine = ({ text }: { text: string }) => {
```
madhuravius, line 9: i think we need to be very careful with this because some log lines aren't formatted this way (i think)

minor suggestion for this unless we never run into an error is to have a fallback that just has a minimally formatted line or indicate we failed in formatting it somehow

--

can be done later as long as this doesn't impact too negatively (i haven't seen this very often so this might just be overly careful behavior on my part)

madhuravius: i think we already do this based on line 11, disregard, misread

src/deploy/environment/index.ts
neurosnap: this selector is no longer needed now that we have a fetch ops by org endpoint

src/ui/shared/log-viewer.tsx
```diff
@@ -0,0 +1,81 @@
...
+export const LogViewer = ({ op }: { op: DeployOperation }) => {
+  const wrapper = "font-mono bg-black p-2 rounded-lg text-black-200";
+  const action = fetchOperationLogs({ id: op.id });
+  const loader = useApi(action);
+  const data: string = useSelector((s: AppState) =>
+    selectDataById(s, { id: action.payload.key }),
+  );
+  useEffect(() => {
+    if (op.status === "succeeded" || op.status === "failed") {
+      loader.trigger();
```
madhuravius, line 54: if loader is mutable or trigger is ever not immediately present, might wanna add to dep array

madhuravius (commented):
lgtm, i can approve in the morn if you have a few minutes to discuss activity/index.ts search filtering and possible ways to prevent htat form getting too crazy

i think we should add some tests i didn't see on search
i think we need to test our changes on ftux to ensure onboarding users are not impacted

madhuravius (approved):

PR #127, opened 2023-06-01
chore: added plans page in anticipation of usage

madhuravius:
This is mostly for plans but also includes some tests for signup flow.

---

This does work for now:

image

There are some NaN values (we can default later when we loop back to needing this) but these are hydrated in production apparently (needs verification).

Changing works as expected:

image


src/ui/pages/plans.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+  });
+  it("errors on plans list load failure", async () => {
+    const { TestProvider } = setupIntegrationTest();
+    render(
+      <TestProvider>
+        <PlansPage />
+      </TestProvider>,
+    );
+    server.use(rest.get(`${testEnv.apiUrl}/plans*`, commonFail));
+    expect(await screen.findByText("mock error message")).toBeDefined();
```
neurosnap, line 149: Honestly, this is one of those areas that is really confusing to me and I don't have a great solution.

```
await screen.findByText("mock error message")
```

If this fails then the test will fail, so why do we need an `expect`?

https://testing-library.com/docs/react-testing-library/example-intro#full-example

src/ui/pages/plans.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <div className="flex text-center items-center justify-center mt-4">
+        <div className="max-w-2xl">
+          <p>
+            Your trial has expired, choose a plan to continue or{" "}
+            <Link to={logoutUrl()}>Log Out</Link>
+          </p>
+        </div>
+      </div>
+      {updatePlanLoader.isError || updatePlanLoader.meta?.isWarning ? (
+        <BannerMessages {...updatePlanLoader} />
```
neurosnap, line 46: The entire point of `BannerMessages` was to avoid needing to do all of these checks.  We need to refactor this to be easier to use.

So I'm curious, why are you doing all these conditional checks here?

src/deploy/plan/index.ts
```diff
@@ -0,0 +1,349 @@
...
+  activePlanSelectors.selectById,
+);
+export const { selectTableAsList: selectActivePlansAsList } =
+  activePlanSelectors;
+export const activePlanReducers = createReducerMap(activePlanSlice);
+export const selectFirstActivePlan = createSelector(
+  selectActivePlansAsList,
+  (activePlans) => {
+    if (activePlans.length === 0) {
+      return defaultActivePlan();
```
neurosnap, line 271: `initActivePlan`

src/ui/pages/plans.tsx
madhuravius: because if it's a success we have random checkboxes everywhere on update, can review, maybe i had a missing update or config i'm missing

madhuravius: maybe i can add a optional param to it that's like hideonsuccess or something? 

or a variant of bannermessages that is like errormessages that banner messages extends?

madhuravius: updated

neurosnap (approved):

PR #132, opened 2023-06-07
permissions ui

neurosnap:
Adds the following functions to be used in our app:
- `selectUserHasPerms` (permission business logic)
- `usePermsRequired` (redirects if user doesn't have perms)
- `PermissionGate` (return `null` if user doesn't have perms)
- `PermRequired` (layout that redirects if user doesn't have perms)
- `createButtonPermission` (button generator for perms)
- `ButtonCreate` (disables button if user doesn't have ability to create resources)
- `ButtonDestroy` (disables button if user doesn't have ability to destroy resources)

### DIFF
--- a/src/deploy/permission/selectors.test.ts
+++ b/src/deploy/permission/selectors.test.ts
@@ -1,0 +1,247 @@
+// @vitest-environment node
+import { defaultPermission, selectUserHasPerms } from "./index";
+import { createId } from "@app/mocks";
+import { defaultRole } from "@app/roles";
+import { AppState, DeepPartial, PermissionScope } from "@app/types";
+
+describe("selectUserHasPerms", () => {
+  describe("when user is an `owner`", () => {
+    it("should grant full read/write access to any account", () => {
+      const role = defaultRole({
+        id: `${createId()}`,
+        name: "Owners",
+        type: "owner",
+      });
+      const envId = `${createId()}`;
+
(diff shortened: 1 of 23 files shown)


src/ui/shared/activity.tsx
```diff
@@ -11,9 +11,9 @@ import {
   cancelDatabaseOpsPoll,
   cancelEnvOperationsPoll,
   cancelOrgOperationsPoll,
-  fetchAllApps,
-  fetchAllDatabases,
-  fetchAllEnvironments,
+  // fetchAllApps,
```
madhuravius, line 14: can't tell if intended (guessing they are unused right now) and so you're leaving it?

src/deploy/permission/selectors.test.ts
```diff
@@ -0,0 +1,247 @@
+// @vitest-environment node
```
madhuravius, line 1: 🤔 do we need to do this for other files or is this something we can set up in a config?

neurosnap: We set the default for jsdom since most of our tests will be react-based.  For one-offs like this we have to opt-in to node to make the tests run faster.

madhuravius: good to know, mainly for pure js related tests or things that don't need access to the dom

PR #136, opened 2023-06-08
chore: more granular error messages 

madhuravius:
If codescan result is python, we show the following for users that might run into common python issues with builder:

image

---

- [x] consolidate items

QA on view Dockerfile if avail

• Banner
• Toggle Control Label (Hide Dockerfile | Show Dockerfile)
• Code Snippet
• Divider line

image

image

### DIFF
--- a/src/ui/pages/create-project-git.tsx
+++ b/src/ui/pages/create-project-git.tsx
@@ -1106,20 +1106,60 @@ const CodeScanInfo = ({
     );
   }
 
-  return (
-    <Banner variant="info">
+  const commonHelpText = (
+    <>
       <span>
-        Your code is missing a Dockerfile to deploy. We will try to generate one
-        for you. We recommend adding a{" "}
-      </span>
-      <ExternalLink
-        href="https://www.aptible.com/docs/dockerfile"
-        variant="info"
-      >
-        Dockerfile
-      </ExternalLink>
+          We recommend adding a{" "}
+          </span>
+          <ExternalLink
+            href="https://www.aptible.com/docs/dockerfile"
+            variant="info"
+          >
+            Dockerfile
+          </ExternalLink>
       <span> to your repo, commit it, and push your code.</span>
-    </Banner>
+    </>
+  )
+
+  if (codeScan.languages_detected?.includes("python")) {
+    return (
+      <Banner variant="info">
+        <div className="ml-2">
+          <p>
+            <strong>We have detected a Python application that does not contain a Dockerfile.</strong>
+          </p>
+          <p className="my-2">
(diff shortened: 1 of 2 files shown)


src/ui/pages/create-project-git.tsx
```diff
@@ -1106,20 +1106,69 @@ const CodeScanInfo = ({
...
         variant="info"
       >
         Dockerfile
       </ExternalLink>
       <span> to your repo, commit it, and push your code.</span>
-    </Banner>
+    </>
+  );
+
+  if (codeScan.languages_detected?.includes("python")) {
```
neurosnap, line 1122: No need for optional chaining here, right?

src/ui/pages/create-project-git.tsx
```diff
@@ -1106,20 +1106,69 @@ const CodeScanInfo = ({
...
+            <strong>
+              We have detected a Python application that does not contain a
+              Dockerfile.
+            </strong>
+          </p>
+          <p className="my-2">
+            There will need to be a{" "}
+            <pre className="inline">requirements.txt</pre> and/or{" "}
+            <pre className="inline">pyproject.toml</pre> in the root directory
+            of your app We suggest ensuring the following:
```
neurosnap, line 1136:
```suggestion
            of your app. We suggest ensuring the following:
```

src/ui/pages/create-project-git.tsx
```diff
@@ -1106,20 +1106,69 @@ const CodeScanInfo = ({
...
         Dockerfile
       </ExternalLink>
       <span> to your repo, commit it, and push your code.</span>
-    </Banner>
+    </>
+  );
+
+  if (codeScan.languages_detected?.includes("python")) {
+    return (
+      <Banner variant="info">
```
neurosnap, line 1124: This banner feels big and I almost feel like we need to upgrade this to `warning` since there is something actionable here for the user.  I wonder if multiple banners would help?

neurosnap (changes requested):

src/ui/pages/create-project-git.tsx
madhuravius: It feels HUGE and I agree, I did this initially as a warning (ugly), and I did this as multiple banners, also ugly. 

I THINK during our pair/reviewathon today for other things we're planning (the sequence of 1 pointers for all the read views), we can do this then w/ Joe

joeclayallday:
Screen Shot 2023-06-08 at 9 48 33 AM

madhuravius:
Opting to do this right now:

image

neurosnap (approved):

PR #138, opened 2023-06-08
feat: added notification banner of active status / some editable settings

madhuravius:
This applies on both apps and databases.

Need to talk through some quirks/issues detected:

* how do users dismiss this?
* how old do things have to be for us to show it?

- [x] make sure name is matching not the word delete for confirmation
- [x] 404 issues on completion of deprov needs to get chatted w/ as a group (using custom thunks for now)
- [x] check if apps need to be reloaded
- [x] ~unit tests (for another PR)~
- [x] discuss some of the dynamism that needs to be resolved

---

Failure: 

image

Success:

image

In progress:

image

Warnings on name change:

image


src/deploy/log-drain/index.ts
```diff
@@ -64,6 +69,15 @@ const initLogDrain = defaultDeployLogDrain();
 const must = mustSelectEntity(initLogDrain);
 export const selectLogDrainById = must(selectors.selectById);
 export const { selectTableAsList: selectLogDrainsAsList } = selectors;
+export const selectLogDrainsByEnvId = createSelector(
+  selectLogDrainsAsList,
+  (_: AppState, props: { envId: string }) => props.envId,
+  (logDrains, envId) => {
+    return logDrains
+      .filter((logDrain) => logDrain.environmentId === envId)
+      .sort((a, b) => a.id.localeCompare(b.id));
```
neurosnap, line 78: i'd probably sort by `createdAt` or `updatedAt`

src/deploy/metric-drain/index.ts
```diff
@@ -62,6 +67,15 @@ const initMetricDrain = defaultDeployMetricDrain();
 const must = mustSelectEntity(initMetricDrain);
 export const selectMetricDrainById = must(selectors.selectById);
 export const { selectTableAsList: selectMetricDrainsAsList } = selectors;
+export const selectMetricDrainsByEnvId = createSelector(
+  selectMetricDrainsAsList,
+  (_: AppState, props: { envId: string }) => props.envId,
+  (metricDrains, envId) => {
+    return metricDrains
+      .filter((metricDrain) => metricDrain.environmentId === envId)
+      .sort((a, b) => a.id.localeCompare(b.id));
```
neurosnap, line 76: Same here

src/ui/pages/db-detail-settings.tsx
```diff
@@ -9,34 +11,67 @@ import {
...
 
   useEffect(() => {
     setHandle(database.handle);
   }, [database.id]);
 
   const onSubmitForm = (e: SyntheticEvent) => {
     e.preventDefault();
+
+    setIsUpdating(true);
+    setTimeout(() => {
```
neurosnap, line 62: Is this to show some information to the user before the request finishes?

src/ui/shared/active-operation-notice.tsx
```diff
@@ -0,0 +1,100 @@
...
+      return selectLatestOpByDatabaseId(s, { dbId: resourceId });
+    }
+  });
+
+  if (!operation) {
+    return null;
+  }
+
+  const bannerStatus: StatusVariant =
+    operation.status === "succeeded"
```
neurosnap, line 40: I would suggest turning this into a function for readability.

src/ui/shared/active-operation-notice.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const ActiveOperationNotice = ({
+  resourceId,
+  resourceType,
+  heartbeat,
+}: {
+  resourceId: string;
+  resourceType: Extract<ResourceType, "app" | "database">;
+  heartbeat?: Date;
+}) => {
+  const operation = useSelector((s: AppState) => {
```
neurosnap, line 27: What do you think about creating two separate components that we render based on the `resourceType`?  This would get rid of the need to have a branch inside one of our selectors.

Conversely, you could put this logic directly inside a more general selector and pass in the `resourceType`.

src/ui/shared/banner.tsx
```diff
@@ -28,10 +28,12 @@ export const Banner = ({
   children,
   className = "",
   variant = "info",
+  iconOverride,
 }: {
   children: React.ReactNode;
   className?: string;
   variant?: StatusVariant;
+  iconOverride?: React.ReactNode;
```
neurosnap, line 36: Hmm, why do we need this override?  Feels like a smell to me.

src/ui/hooks/use-poller.ts
```diff
@@ -0,0 +1,30 @@
...
+  cancel,
+  heartbeatFunc,
+}: {
+  action: Action;
+  cancel: Action;
+  heartbeatFunc?: (date: Date) => void;
+}) => {
+  const dispatch = useDispatch();
+
+  useEffect(() => {
```
neurosnap, line 16: This seems unrelated to polling, what do you think about making it a separate hook?

neurosnap (commented):
Overall solid work.  I don't think there's much that would block this from going in as-is.

I'm a little fuzzy on the business requirements, but I thought we were only going to show deprovisioning and not provisioning/deploy?  Did I remember that wrong?  If we just did deprovisioning then we wouldn't need to worry about dismissing the banner, right?

src/ui/pages/db-detail-settings.tsx
madhuravius: i should've left a todo. 

the request happens SO fast that it's weird, it feels like it didn't even register

src/ui/shared/banner.tsx
madhuravius: because i want to override what's in a banner for a spinning in progress icon under specific situations

madhuravius: i also wanted it unset in some cases as required, i think, we can break this up as well into multiple attrs

src/ui/shared/active-operation-notice.tsx
madhuravius: i initially did that (two components) and it was VERY repetitive. 

The only fork is the common item, but i think you have a good idea, maybe I should make a SINGLE selector and return a slightly different generic type on common fields we can iterate over instead of combining them.

src/ui/shared/banner.tsx
madhuravius: switched to a show/hide

src/ui/hooks/use-poller.ts
madhuravius: i can do that sure, having another hook that's just a useHeartbeat or something

madhuravius: `src/ui/hooks/use-interval.ts` does that already so gonna use that oops...

src/ui/pages/db-detail-settings.tsx
madhuravius: this was slow enough when testing that i dropped it

src/ui/shared/activity.tsx
```diff
@@ -215,6 +218,8 @@ export function ActivityByOrg({ orgId }: { orgId: string }) {
         onChange={onChange}
         isLoading={loader.isLoading}
         search={search}
+        title="Activity"
+        description="Realtime dashboard of your organization's operations in the last week."
```
madhuravius, line 222: might wanna chat w/ joe on these he really wants them gone

joeclayallday:
Looks good! Only change I see is remove the space before the ellipsis here:
Screen Shot 2023-06-08 at 11 08 46 PM

And yeah, I don't think we need the activity page description always showing, it will speak for itself 👍

neurosnap (dismissed):

neurosnap (approved):

PR #147, opened 2023-06-09
design(activity): ux tweaks

neurosnap:


madhuravius (approved):
feels like we should just do a toUpper JS equivalent

PR #150, opened 2023-06-13
chore: removed or disabled inputs where there is no actionable outcome

madhuravius:
Either via disabled attributes, comments or deletions, non-actionable buttons/inputs are now disabled


almathew (approved):
I get the commenting out here, but I'm approving assuming you actually commented out everything (and without testing to confirm that)

PR #158, opened 2023-06-20
Properly wipe FE store after certain operations

neurosnap:


src/deploy/app/index.ts
```diff
@@ -158,8 +160,20 @@ export const selectFirstAppByEnvId = createSelector(
   (apps) => apps[0] || initApp,
 );
 
-export const selectAppsForTable = createSelector(
+export const selectAppsByOrgAsList = createSelector(
   selectAppsAsList,
+  selectEnvironmentsByOrg,
+  selectOrganizationSelectedId,
+  (apps, envs) => {
+    return apps.filter((app) => {
```
madhuravius, line 168: nit: funcitons below don't use returns and just rely on arrow func returns (no brace), not sure why linter isn't handling this

src/test/index.tsx
```diff
@@ -21,10 +22,11 @@ export const setupTestStore = (
   });
 
   middleware.push(...prepared.middleware);
+  const baseReducer = resetReducer(prepared.reducer, persistConfig);
```
madhuravius, line 25: 👌 great so we'll have comparable behavior in tests

src/api/index.ts
```diff
@@ -152,6 +158,29 @@ function* requestAuth(ctx: ApiCtx, next: Next): ApiGen {
...
+  if (!ctx.response) return;
+  if (ctx.response.status === 401) {
+    ctx.actions.push(resetToken());
+  }
+}
+
+function* aborter(ctx: ApiCtx, next: Next) {
+  const signal = yield* select(selectSignal);
+  const aborted = () =>
+    new Promise<void>((resolve) => {
```
madhuravius, line 172: nit/request - did you write this in a totally new way or did you see this somewhere and reference it? if the latter can you include a comment for anyone reading this later?

madhuravius (approved):

PR #159, opened 2023-06-20
fix: properly filter by env (and empty resource tables)

madhuravius:
no app, one db:

image

one app, no db:

image

one app, one db:

image

many dbs:

image


src/deploy/app/index.ts
```diff
@@ -170,6 +170,14 @@ export const selectAppsForTable = createSelector(
       .sort((a, b) => a.handle.localeCompare(b.handle)),
 );
 
+export const selectAppsForTableByEnvironmentId = createSelector(
+  selectAppsForTable,
+  (_: AppState, props: { envId: string }) => props.envId,
+  (apps, envId) => {
+    return apps.filter((app) => app.environmentId === envId);
```
madhuravius, line 177: needed because there's mutated data from parent selector that is not present for the db equiv

joeclayallday (approved):

src/deploy/app/index.ts
neurosnap: If you use this inside the view then it won't have all the grep filtering, right?

You can see how I did the exact same thing here: https://github.com/aptible/app-ui/blob/edda4368cc09b34568c46398372a38182e2b8fb6/src/deploy/activity/index.ts#L44

src/ui/shared/app/app-list.tsx
```diff
@@ -139,9 +151,11 @@ export function AppList({
     setSearch(ev.currentTarget.value);
 
   const apps = useSelector((s: AppState) =>
-    selectAppsForTableSearch(s, {
-      search: searchOverride ? searchOverride : search,
-    }),
+    environmentId
+      ? selectAppsForTableByEnvironmentId(s, { envId: environmentId })
+      : selectAppsForTableSearch(s, {
```
neurosnap, line 156: We should make separate components for this and not branch.

You can see how I did this inside `Activity`:
https://github.com/aptible/app-ui/blob/edda4368cc09b34568c46398372a38182e2b8fb6/src/ui/shared/activity.tsx#L212

Another reason to no use the global `AppList` component is so we can be more surgical with what we fetch.  For an `AppList` scoped to a single `Account` we should use an endpoint that fetches just those apps.

src/ui/shared/db/database-list.tsx
```diff
@@ -99,9 +105,11 @@ export function DatabaseList({
     setSearch(ev.currentTarget.value);
 
   const dbs = useSelector((s: AppState) =>
-    selectDatabasesForTableSearch(s, {
-      search: searchOverride ? searchOverride : search,
-    }),
+    environmentId
```
neurosnap, line 108: Same thing here.

neurosnap (changes requested):

src/deploy/app/index.ts
madhuravius: understood, yeah i'll enforce it's a search result on the view as well (bad mistake on my part)

src/ui/shared/resource-list-view.tsx
```diff
@@ -12,7 +13,7 @@ interface ResourceHeaderProps {
 }
 
 interface GenericResourceListProps {
-  header?: JSX.Element;
+  header?: ReactNode;
```
madhuravius, line 16: https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

this must be because there are situations where the header can be nil/empty apparently from the existing code

madhuravius:
So it looks like we're going to:

* break up apps/dbs on env page
* add a bunch of new data points on those tables
* drop search on the env specific views (and links to the same env)

madhuravius:
@joeclayallday after recent changes:

image

dbs list:

image

apps list:

image

dbs in env:

image

apps in env:

image

joeclayallday (approved):

madhuravius:
@neurosnap  on the note of overly nestedness/lots of complexity on the views for the lists

i'm wrapping up comparable tables with similar layout issues eric, so i can pull a lot of the common elements (placeholder empty table, others) into common shared components and add them to styles later tomorrow evening

neurosnap (approved):

PR #160, opened 2023-06-20
chore: minor QA

madhuravius:
link styling for log buttons:

image

title:
image

latest operation:
image

icon updated (tested w/ store):
image

resource name where possible:
image

---

On the topic of links: Note this whole thing is wrapped weird due to styling issues of needing `px-0` on buttons which will not be honored on ButtonLink due to colliding styles


joeclayallday (approved):

index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Aptible Deploy</title>
+    <title>Aptible</title>
```
neurosnap, line 7: Non-blocker, just noting: We should probably start thinking about adding a react head component like https://github.com/nfl/react-helmet

src/ui/layouts/environment-detail-layout.tsx
```diff
@@ -5,119 +5,150 @@ import { environmentsUrl } from "@app/routes";
...
   AppState,
   DeployEndpoint,
   DeployEnvironment,
+  DeployOperation,
   DeployStack,
 } from "@app/types";
 import cn from "classnames";
 import { useQuery } from "saga-query/react";
 
 const environmentDetailBox = ({
```
neurosnap, line 36: What's the rationale for this formatting?  This is just a react component, it should be treated as such.

src/ui/layouts/environment-detail-layout.tsx
```diff
@@ -5,119 +5,150 @@ import { environmentsUrl } from "@app/routes";
...
-      <div className="flex w-1/1">
-        <div className="flex-col w-1/3">
-          <div className="mt-4">
-            <h3 className="text-base font-semibold text-gray-900">ID</h3>
-            <p>{environment.id}</p>
-          </div>
-          <div className="mt-4">
-            <h3 className="text-base font-semibold text-gray-900">Stack</h3>
-            <p>{stack.name}</p>
+}): React.ReactElement => {
```
neurosnap, line 46: This return type in unnecessary.

src/ui/layouts/environment-detail-layout.tsx
```diff
@@ -5,119 +5,150 @@ import { environmentsUrl } from "@app/routes";
...
+  const userName =
+    latestOperation.userName.length >= 15
+      ? `${latestOperation.userName.slice(0, 15)}...`
+      : latestOperation.userName;
+  return (
+    <div className={cn(tokens.layout["main width"], "py-6 -mt-5 -mb-5")}>
```
neurosnap, line 52: These negative margins are a smell, we should pair on removing them.

src/ui/layouts/environment-detail-layout.tsx
```diff
@@ -5,119 +5,150 @@ import { environmentsUrl } from "@app/routes";
...
-          </div>
-          <div className="mt-4">
-            <h3 className="text-base font-semibold text-gray-900">
-              {environment.totalDatabaseCount} Database
-              {environment.totalDatabaseCount > 0 && "s"}
-            </h3>
-            {environment.databaseContainerCount} container
-            {environment.databaseContainerCount > 0 && "s"} using{" "}
-            {environment.totalDiskSize} GB of disk
+        <div className="flex w-1/1">
```
neurosnap, line 73: `w-1/1` is not a thing as far as I can tell.  What are you trying to accomplish here?

neurosnap (changes requested):

index.html
madhuravius: yeah i spoke to joe about this and that's not a bad idea (there's ways to do this with react router as well (maybe using that lib))

src/ui/layouts/environment-detail-layout.tsx
```diff
@@ -144,6 +175,7 @@ function EnvironmentPageHeader(): React.ReactElement {
       detailsBox={environmentDetailBox({
```
neurosnap, line 175: This should just be a react component.

src/ui/layouts/environment-detail-layout.tsx
madhuravius: IDE was complaining, so I was forced to in this case (inference wasn't working properly), i can try again after resetting local lang serv

src/ui/layouts/environment-detail-layout.tsx
madhuravius: agreed, don't remember doing this, i will check again

src/ui/layouts/environment-detail-layout.tsx
madhuravius: copy pasta and likely mistake, i should search for other occurrences

src/ui/layouts/environment-detail-layout.tsx
```diff
@@ -5,119 +5,150 @@ import { environmentsUrl } from "@app/routes";
...
-          <div className="mt-4">
-            <h3 className="text-base font-semibold text-gray-900">
-              {environment.totalDatabaseCount} Database
-              {environment.totalDatabaseCount > 0 && "s"}
-            </h3>
-            {environment.databaseContainerCount} container
-            {environment.databaseContainerCount > 0 && "s"} using{" "}
-            {environment.totalDiskSize} GB of disk
+        <div className="flex w-1/1">
+          <div className="flex-col w-1/3">
```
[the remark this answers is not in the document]
madhuravius: we can drop it, but it was from multicol support here (which we can use with flex too):

image

src/ui/layouts/environment-detail-layout.tsx
```diff
@@ -5,119 +5,150 @@ import { environmentsUrl } from "@app/routes";
...
-          <div className="mt-4">
-            <h3 className="text-base font-semibold text-gray-900">ID</h3>
-            <p>{environment.id}</p>
-          </div>
-          <div className="mt-4">
-            <h3 className="text-base font-semibold text-gray-900">Stack</h3>
-            <p>{stack.name}</p>
+}): React.ReactElement => {
+  const userName =
+    latestOperation.userName.length >= 15
```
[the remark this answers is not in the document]
madhuravius: i copy pastaed this from an existing activity page, ill update it in both from your js handles it properly suggestion (since it seems easiest to read)

src/ui/layouts/environment-detail-layout.tsx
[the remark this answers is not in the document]
madhuravius: we're coming back to this

neurosnap (approved):

PR #164, opened 2023-06-22
feat(stacks): stack list

neurosnap:


src/ui/pages/stacks.tsx
```diff
@@ -0,0 +1,136 @@
...
+    selectDatabasesCountByStack(s, { stackId: stack.id }),
+  );
+
+  return (
+    <tr>
+      <Td>
+        <div className="flex items-center">
+          <img
+            src={
+              stackType === "dedicated"
```
madhuravius, line 52: urls are slightly different for this now, they are prefixed in a path (we can also flatten them to top level but i was worried about collisions)

src/deploy/database/index.ts
```diff
@@ -288,6 +288,23 @@ export const selectDatabasesByEnvId = createSelector(
...
+    return dbs.filter((db) => {
+      const env = findEnvById(envs, { id: db.environmentId });
+      return env.stackId === stackId;
+    });
+  },
+);
+
+export const selectDatabasesCountByStack = createSelector(
+  selectDatabasesByStack,
+  (dbs) => dbs.length,
```
madhuravius, line 305: should i update all the other counts i've been doing with this? i just haven't been doing this yet

neurosnap: The main reason I did this was for memoization, not sure if it really helps that much so I wouldn't worry about it.

madhuravius (approved):

PR #166, opened 2023-06-23
update-readme-jc

joeclayallday:
**Update our readme with clear instructions for contributors.**

I'm hosting the image of the site here and referencing the URL in the readme.

Screen Shot 2023-06-23 at 10 59 47 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@ Aptible's PaaS automates the work of provisioning, managing, and scaling infrast
 app-ui requires node `v18.15.x` and yarn `v2+` or later.
 
 **Add Environment Variables**
-
 in root of app-ui create a `.env` file with these environment variables:
 ```bash
 VITE_AUTH_URL="https://auth.aptible.com"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,54 +1,73 @@
...
+**1. Install Node and Yarn**
+<br>app-ui requires node `v18.15.x` and yarn `v2+` or later.
 
+**2. Add Environment Variables**
+<br>Create a `.env` file with these variables in the root of app-ui:
 ```bash
-yarn
+VITE_AUTH_URL="https://auth.aptible.com"
+VITE_API_URL="https://api.aptible.com"
+VITE_ORIGIN="nextgen"
```
grrowe, line 20: I would add a note that you can also just have these env variables set with a `.env` file. 
Also, another note to tell users that these variables point to which API/Auth to use, like they could use the sbx.main ones as well.

madhuravius: there should be an example - https://github.com/aptible/app-ui/blob/main/.env.example

they can use that as a starting point

joeclayallday:
Updated Mockup with revisions:
Screen-Shot-2023-06-23-at-12 51 22-PM

README.md
joeclayallday: Added in these tweaks, thanks!

madhuravius: i meant they can use the example file, like:

```sh
cp .env.example .env
```

and theyll have a file ready to go they just gotta tweak

madhuravius:
LGTM, left minor note to clarify

grrowe (approved):

PR #170, opened 2023-06-23
reactor: detail header components

neurosnap:
Screenshot 2023-06-26 at 12 11 56 PM

### DIFF
--- a/src/ui/pages/stack-detail.tsx
+++ b/src/ui/pages/stack-detail.tsx
@@ -1,0 +1,8 @@
+import { Navigate, useParams } from "react-router";
+
+import { stackDetailEnvsUrl } from "@app/routes";
+
+export const StackDetailPage = () => {
+  const { id = "" } = useParams();
+  return <Navigate to={stackDetailEnvsUrl(id)} replace />;
+};
--- a/src/ui/shared/activity.tsx
+++ b/src/ui/shared/activity.tsx
@@ -206,31 +206,29 @@ function ActivityTable({
   };
 
   return (
-    <div className="my-4">
-      <ResourceListView
-        header={resourceHeaderTitleBar()}
-        tableHeader={
-          <TableHead
-            headers={[
-              "Resource",
-              "Status",
-              "Operation Type",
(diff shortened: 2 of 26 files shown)


src/ui/shared/tabs.tsx
```diff
@@ -43,7 +43,7 @@ export const Tabs = ({ tabs }: { tabs: TabItem[] }) => {
       {tabs.map((tab) => (
         <Tab
           key={tab.name}
-          isActive={tab.current}
+          isActive={!!tab.current}
```
madhuravius, line 46: nit - shouldn't need double bangs 🤔 (you told me this some time ago)

i prefer this sometimes, but we should be consistent about it

neurosnap: `tab.current` can now be undefined so it's not always a boolean.

src/ui/pages/stack-detail.tsx
```diff
@@ -0,0 +1,8 @@
+import { Navigate, useParams } from "react-router";
+
+import { stackDetailEnvsUrl } from "@app/routes";
+
+export const StackDetailPage = () => {
+  const { id = "" } = useParams();
+  return <Navigate to={stackDetailEnvsUrl(id)} replace />;
```
madhuravius, line 7: 🤔 until this is ready i'm guessing. we should probably do this more often

src/ui/shared/activity.tsx
```diff
@@ -206,31 +206,29 @@ function ActivityTable({
...
+            "Status",
+            "Operation Type",
+            "Environment",
+            "User",
+            "Last Updated",
+            "Actions",
+          ]}
+        />
+      }
+      tableBody={
```
madhuravius, line 224: table body should already have a parent div imo and this should contain the rows and be explicit it expects either:

`[]OpListRow` or `children` (ReactNode or something?) so this should just work OOB that way

madhuravius (approved):
LGTM, nearly exactly what i had in mind when planning on returning to this work

PR #186, opened 2023-07-10
[Merge after Charts] Change DB Tab Order

joeclayallday:
Make Databases be the first tab and link destination from DB list table

Screen Shot 2023-07-12 at 12 07 42 PM


madhuravius:
2nd screenshot is invalid config on local (chatted on slack rq)

i don't think we should do this change until we have charts merged though and other backend changes merged up

joeclayallday:
I got this fixed by changing my .env file and now i see this:

Screen Shot 2023-07-10 at 10 55 47 AM

PR #208, opened 2023-07-17
feat: omnisearch

neurosnap:
https://github.com/aptible/app-ui/assets/1940365/cd31e86b-4def-4cba-8a4a-6fb7197c054b

### DIFF
--- a/src/search/index.ts
+++ b/src/search/index.ts
@@ -1,0 +1,109 @@
+import {
+  selectAppsAsList,
+  selectDatabasesAsList,
+  selectEnvironmentsAsList,
+  selectStacksAsList,
+} from "@app/deploy";
+import {
+  AppState,
+  DeployApp,
+  DeployDatabase,
+  DeployEnvironment,
+  DeployStack,
+} from "@app/types";
+import { createSelector } from "@reduxjs/toolkit";
+
+interface AbstractResourceItem {
+  id: string;
+  handle: string;
+}
+
+export interface StackItem extends AbstractResourceItem {
+  type: "stack";
+  data: DeployStack;
+}
+
+export interface EnvItem extends AbstractResourceItem {
+  type: "environment";
+  data: DeployEnvironment;
+}
+
+export interface AppItem extends AbstractResourceItem {
+  type: "app";
+  data: DeployApp;
+}
+
+export interface DbItem extends AbstractResourceItem {
+  type: "database";
+  data: DeployDatabase;
+}
+
+export type ResourceItem = StackItem | EnvItem | AppItem | DbItem;
+
+export const selectResourcesAsList = createSelector(
+  selectStacksAsList,
+  selectEnvironmentsAsList,
+  selectAppsAsList,
+  selectDatabasesAsList,
+  (stacks, envs, apps, dbs): ResourceItem[] => {
+    const resources: ResourceItem[] = [];
+    stacks.forEach((stack) => {
+      resources.push({
+        type: "stack",
+        id: stack.id,
+        handle: stack.name.toLocaleLowerCase(),
+        data: stack,
+      });
+    });
+
+    envs.forEach((env) => {
+      resources.push({
+        type: "environment",
+        id: env.id,
+        handle: env.handle.toLocaleLowerCase(),
+        data: env,
+      });
+    });
+
+    apps.forEach((app) => {
+      resources.push({
+        type: "app",
+        id: app.id,
+        handle: app.handle.toLocaleLowerCase(),
+        data: app,
+      });
+    });
+
+    dbs.forEach((db) => {
+      resources.push({
+        type: "database",
+        id: db.id,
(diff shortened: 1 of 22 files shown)


src/search/index.ts
```diff
@@ -0,0 +1,184 @@
...
+      if (dateB === dateA) {
+        return b.count - a.count;
+      }
+
+      return dateB - dateA;
+    });
+  },
+);
+
+export const selectResourcesByMostVisited = createSelector(
```
madhuravius, line 140: neat section, not quite sure how it's used yet, still reading, but now you have my attention

src/ui/pages/search.tsx
```diff
@@ -0,0 +1,195 @@
...
+export const SearchPage = () => {
+  const [params, setParams] = useSearchParams();
+  const search = params.get("search") || "";
+  const resources = useSelector((s: AppState) =>
+    selectResourcesForSearch(s, { search }),
+  );
+  const onChange = (ev: React.ChangeEvent<HTMLInputElement>) =>
+    setParams({ search: ev.currentTarget.value });
+  const curLimit = Math.min(resources.length, SEARCH_DISPLAY_LIMIT);
+  const resultText = `Displaying ${curLimit} of ${resources.length} results`;
```
madhuravius, line 157: personal pref - if we have em all i don't see why we don't just show it all and make it overflow capable, people can scroll to get whatever they want (unless the desire is to limit this on the backend later)

src/ui/pages/search.tsx
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  if (resource.type === "app") {
+    return <AppResource resource={resource} />;
+  }
+
+  return <DbResource resource={resource} />;
+};
+
+const SEARCH_DISPLAY_LIMIT = 15;
```
madhuravius, line 115: we should probably keep this in a const somewhere, i can see us reusing this

src/ui/shared/application-sidebar.tsx
```diff
@@ -82,6 +84,22 @@ export const ApplicationSidebar = () => {
...
+            <div
+              className="border border-gray-200 bg-white rounded-lg px-2 py-1 flex items-center justify-between h-[36px] cursor-pointer hover:bg-black-50"
+              onClick={() => navigate(searchUrl())}
+              onKeyUp={() => navigate(searchUrl())}
+            >
+              <div className="flex items-center">
+                <IconSearch color="#888C90" variant="sm" />{" "}
+                <div className="ml-2 text-black-300">Search...</div>
+              </div>
+              {/* <div className="text-black-300">&#8984; K</div> */}
```
madhuravius, line 98: reminder

src/ui/shared/db/database-list.tsx
```diff
@@ -28,25 +28,32 @@ import { capitalize } from "@app/string-utils";
...
 type DatabaseCellProps = { database: DeployDatabase };
 
+export const DatabaseItemView = ({
+  database,
+}: { database: DeployDatabase }) => {
+  return (
+    <div className="flex">
+      <Link to={databaseEndpointsUrl(database.id)} className="flex">
+        <img
+          src={`/database-types/logo-${database.type}.png`}
```
madhuravius, line 38: TODO for us - we should probably make these const strings at some point

src/ui/shared/stack.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { DeployStack } from "@app/types";
+import { Link } from "react-router-dom";
+
+export const StackItemView = ({ stack }: { stack: DeployStack }) => {
+  const stackType = getStackType(stack);
+  return (
+    <div className="flex items-center">
+      <img
+        src={
+          stackType === "dedicated"
```
madhuravius, line 12: same as above, todo for us

src/ui/pages/search.tsx
neurosnap: We want to limit the results because a) we want them to search for what they want and b) if we overload the page with elements it'll get sluggish.

src/search/index.ts
neurosnap: Everytime a user goes to a resource (e.g. the detail header component is loaded) we track it as a view count.  Then we use this selector to sort by view count for each resource the user visited.

src/ui/shared/application-sidebar.tsx
neurosnap: Keeping this for now, we'll want to add it back when we have the omnisearch fully functional with keeb shortcut.

src/search/index.ts
madhuravius: this is fine but potentially misleading (probably YOUR most viewed equvialent language is better for now), but lgtm

madhuravius (approved):
LGTM - we have some items to discuss as follow up but this is a great starting addition 🥳

PR #216, opened 2023-07-18
feat: added charts functionality 

madhuravius:
* moved poller into separate component
* moved each chart into separate component with selector pulling data
* moved each table into separate component with selector pulling table data
* lazy loads chart components
* day/month/week selectors for data

---

App charts

image

App table data

image

Db charts

image

Db table data

image

---

Not impactful from footprint perspective (should be pretty cached and those components should load separately entirely from the rest of the app):

image

Builds and runs fine locally:

image

image


src/ui/pages/app-detail-service.tsx
```diff
@@ -35,15 +45,38 @@ export function AppDetailServicePage() {
...
+        />
+        <MetricsHorizonControls
+          viewHorizon={viewHorizon}
+          setViewHorizon={setViewHorizon}
+        />
+      </div>
+      <LoadResources query={query} isEmpty={false}>
+        {containers.map((container) => (
+          <div className="my-4" key={container.id}>
+            {viewTab === "chart" ? (
```
madhuravius, line 62: can probably switch this to routes later at some point, at which point i think the dates may be associated with that but NOT that view tabs

src/ui/shared/metrics-controls.tsx
```diff
@@ -0,0 +1,115 @@
...
+export type MetricTabTypes = "table" | "chart";
+export type MetricHorizons = "1h" | "1d" | "1w";
+export type Dataset = {
+  label?: string;
+  pointRadius?: number;
+  pointHoverRadius?: number;
+  data: number[];
+};
+
+export const processDataSeries = ({
```
madhuravius, line 13: could be moved or we can just change this file into a generic catchall for data manipulation/handling metrics

src/ui/pages/db-detail-metrics.tsx
```diff
@@ -34,14 +44,35 @@ export function DatabaseMetricsPage() {
...
+              viewHorizon={viewHorizon}
+            />
+          )}
         </div>
       ))}
     </LoadResources>
   );
 }
+
+export const Component = DatabaseMetricsPage;
```
madhuravius, line 78: needed IF lazyloaded

src/app/router.tsx
```diff
@@ -324,13 +324,14 @@ export const appRoutes: RouteObject[] = [
...
                 element: <AppDetailServicesPage />,
               },
-              {
-                path: routes.APP_SERVICES_PATH,
-                element: <AppDetailServicesPage />,
-              },
               {
                 path: routes.APP_SERVICE_PATH,
                 element: <AppDetailServicePage />,
+                // lazy: () => import("@app/ui/pages/app-detail-service"),
```
madhuravius, line 330: attempt leaving this in after changing how hooks behave

src/ui/pages/app-detail-service.tsx
madhuravius: make each chart a separate component (instead of looping through them this way)

src/ui/pages/db-detail-metrics.tsx
madhuravius: lazy loading works, so i'm leaving it in

src/ui/shared/metrics-controls.tsx
madhuravius: this should be removable

src/metric-tunnel/index.ts
```diff
@@ -1,10 +1,258 @@
...
+  datasets?: Dataset[];
+};
+export type FlatTableOfMetricsData = {
+  [metric: string]: (number | string)[];
+  time: string[];
+};
+
+const chartTitles: { [key: string]: string[] } = {
+  Memory: ["memory_all"],
+  CPU: ["cpu_pct"],
```
madhuravius, line 29: i should leave a note on part 2 but this is done so we can add more series to the same chart (limits, recommendations etc)

src/metric-tunnel/index.ts
```diff
@@ -1,10 +1,258 @@
...
+  (_: AppState, p: { containerId: string }) => p.containerId,
+  (_: AppState, p: { metricNames: string[] }) => p.metricNames,
+  (_: AppState, p: { metricHorizon: MetricHorizons }) => p.metricHorizon,
+  (
+    containerMetrics,
+    containerId,
+    metricNames,
+    metricHorizon,
+  ): ChartToCreate => {
+    let title = "";
```
madhuravius, line 178: in pt2 - i should comment/explain this, hard to read

src/ui/pages/app-detail-service.tsx
```diff
@@ -8,42 +8,112 @@ import {
...
+            serviceId: service.id,
+          }),
+        );
+      }),
+    );
+    if (actions.length === 0) {
+      return;
+    }
+    dispatch(batchActions(actions));
+  }, [service.id, containers.length, metricHorizon]);
```
neurosnap, line 69: I would use `containers` instead of just the length.

neurosnap (approved):
One comment, otherwise lgtm!

src/ui/pages/app-detail-service.tsx
```diff
@@ -8,42 +8,117 @@ import {
...
     selectContainersByReleaseIdByLayerType(s, {
-      layers: ["app", "database"],
+      layers: layersToSearchForContainers,
       releaseId: service.currentReleaseId,
     }),
   );
 
-  const dataToFetch = ["cpu_pct", "la", "memory_all"];
+  const containerIds = [...containers]
+    .sort()
```
neurosnap, line 52: This sort won't do anything since `containers` is an object, right?

src/ui/pages/db-detail-metrics.tsx
```diff
@@ -9,39 +15,115 @@ import {
...
     selectContainersByReleaseIdByLayerType(s, {
-      layers: ["app", "database"],
+      layers: ["database"],
       releaseId: service.currentReleaseId,
     }),
   );
 
-  const dataToFetch = ["cpu_pct", "la", "memory_all", "iops", "fs"];
+  const containerIds = [...containers]
+    .sort()
```
neurosnap, line 49: Same thing here

neurosnap (approved):

src/ui/pages/app-detail-service.tsx
madhuravius: the array itself will be mutated, this returns an idempotent id chain for me as a result

madhuravius: this potentially has no impact (as the containers should be presorted on the way in, but i'm not trusting it here), im merging but can fix in followup if desired

you're right, the underlying containers are unaffected, but the order of the array itself could be if the selector was not idempotent

PR #218, opened 2023-07-19
chore: fixed things that caused warnings

madhuravius:
before: 

image

after (we can probably squeeze more deps into that hash as well for better caching for people at some point - i just wanted the warnings to go away for now):

image

The only error that's missing is we omit source maps and i think some transforms require it on local


tailwind.config.cjs
```diff
@@ -3,7 +3,7 @@ const flex = defaultTheme.flex;
 
 /** @type {import('tailwindcss').Config} */
 module.exports = {
-  purge: ["./src/**/*.{ts,tsx}", "./index.html"],
```
madhuravius, line 6: https://tailwindcss.com/docs/upgrade-guide#configure-content-sources

tailwind.config.cjs
```diff
@@ -144,6 +144,5 @@ module.exports = {
   plugins: [
     require("@tailwindcss/typography"),
     require("@tailwindcss/forms"),
-    require("@tailwindcss/line-clamp"),
```
madhuravius, line 147:
```
warn - As of Tailwind CSS v3.3, the `@tailwindcss/line-clamp` plugin is now included by default.
warn - Remove it from the `plugins` array in your configuration to eliminate this warning.
```

vite.config.mjs
```diff
@@ -5,6 +5,26 @@ import { defineConfig } from "vitest/config";
 // https://vitejs.dev/config/
 export default defineConfig(() => {
   return {
+    build: {
+      rollupOptions: {
+        output: {
+          manualChunks: {
```
madhuravius, line 11: for size warning or we can omit this

neurosnap (approved):

PR #222, opened 2023-07-20
feat: create standalone app

neurosnap:
https://github.com/aptible/app-ui/assets/1940365/af4b3a7f-c85b-47c8-8bb2-d4cb30e353a6

### DIFF
--- a/src/deploy/index.ts
+++ b/src/deploy/index.ts
@@ -20,3 +20,4 @@ export * from "./release";
 export * from "./container";
 export * from "./vpc_peer";
 export * from "./vpn_tunnel";
+export * from "./code-scan-result";
--- a/src/routes/index.ts
+++ b/src/routes/index.ts
@@ -135,6 +135,9 @@ export const environmentBackupsUrl = (id: string) =>
 export const ENVIRONMENT_SETTINGS_PATH = "/environments/:id/settings";
 export const environmentSettingsUrl = (id: string) =>
   `/environments/${id}/settings`;
+export const ENVIRONMENT_CREATE_APP_PATH = "/environments/:id/app";
+export const environmentCreateAppUrl = (id: string) =>
+  `/create/name?environment_id=${id}`;
 
 export const ERROR_PATH = "/error";
 
--- a/src/test/index.tsx
+++ b/src/test/index.tsx
@@ -1,6 +1,6 @@
 import { prepareStore } from "@app/fx";
 import { Store, configureStore } from "@reduxjs/toolkit";
-import { waitFor } from "@testing-library/react";
+import { act, waitFor } from "@testing-library/react";
 import { Provider } from "react-redux";
 import { RouteObject, RouterProvider, createMemoryRouter } from "react-router";
 import { REHYDRATE } from "redux-persist";
@@ -128,3 +128,13 @@ export const waitForToken = (store: Store<AppState>) => {
     }
   });
 };
+
+export const sleep = (n: number) =>
+  act(
+    () =>
+      new Promise<void>((resolve) => {
+        setTimeout(() => {
+          resolve();
+        }, n);
+      }),
+  );
--- a/src/ui/hooks/use-latest-code-results.ts
+++ b/src/ui/hooks/use-latest-code-results.ts
@@ -1,0 +1,21 @@
+import { fetchAppOperations, selectLatestSucceessScanOp } from "@app/deploy";
+import {
+  DeployCodeScanResponse,
+  fetchCodeScanResult,
+} from "@app/deploy/code-scan-result";
+import { AppState } from "@app/types";
+import { useSelector } from "react-redux";
+import { useCache, useQuery } from "saga-query/react";
+
+export const useLatestCodeResults = (appId: string) => {
+  const appOps = useQuery(fetchAppOperations({ id: appId }));
+  const scanOp = useSelector((s: AppState) =>
+    selectLatestSucceessScanOp(s, { appId }),
+  );
+
+  const codeScan = useCache<DeployCodeScanResponse>(
+    fetchCodeScanResult({ id: scanOp.codeScanResultId }),
+  );
+
+  return { scanOp, codeScan, appOps };
+};
--- a/src/ui/pages/create-project-git-settings.tsx
+++ b/src/ui/pages/create-project-git-settings.tsx
@@ -1,0 +1,956 @@
+import { Reducer, useEffect, useReducer, useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
+import { useNavigate, useParams } from "react-router";
+import { useSearchParams } from "react-router-dom";
+import { selectLoaderById } from "saga-query";
+import { useQuery } from "saga-query/react";
+
+import {
+  fetchAllDatabaseImages,
+  fetchApp,
+  fetchConfiguration,
+  fetchDatabasesByEnvId,
+  fetchServiceDefinitionsByAppId,
+  selectAppById,
+  selectAppConfigById,
+  selectDatabaseImagesAsList,
+  selectDatabasesByEnvId,
+  selectServiceDefinitionsByAppId,
+} from "@app/deploy";
+import { DeployCodeScanResponse } from "@app/deploy";
+import {
+  DB_ENV_TEMPLATE_KEY,
+  DbCreatorProps,
+  TextVal,
+  deployProject,
+  getDbEnvTemplateValue,
+} from "@app/projects";
+import {
+  createProjectGitPushUrl,
+  createProjectGitStatusUrl,
+} from "@app/routes";
+import { AppState, DeployDatabaseImage } from "@app/types";
+
+import { useEnvOpsPoller, useLatestCodeResults, useProjectOps } from "../hooks";
+import {
+  Banner,
+  Box,
+  Button,
+  ExternalLink,
+  FormGroup,
+  IconChevronDown,
+  IconChevronUp,
+  IconPlusCircle,
+  Input,
+  Loading,
+  PreCode,
+  ProgressProject,
+  Select,
+  SelectOption,
+  tokens,
+} from "../shared";
+
+const trim = (t: string) => t.trim();
+const parseText = <
+  M extends { [key: string]: unknown } = { [key: string]: unknown },
+>(
+  text: string,
+  meta: () => M,
+): TextVal<M>[] =>
+  text
+    .split("\n")
+    .map(trim)
+    .map((t) => {
+      // sometimes the value can contain an "=" so we need to only
+      // split the first "=", (e.g. SECRET_KEY=1234=)
+      // https://stackoverflow.com/a/54708145
+      const [key, ...values] = t.split("=").map(trim);
+      const value = Array.isArray(values) ? values.join("=") : values;
+      return {
+        key,
+        value,
+        meta: meta(),
+      };
+    })
+    .filter((t) => !!t.key);
+
+interface ValidatorError {
+  item: TextVal;
+  message: string;
+}
+
+interface DbValidatorError {
+  message: string;
+  item: DbCreatorProps;
+}
+
+const validateDbs = (items: DbCreatorProps[]): DbValidatorError[] => {
+  const errors: DbValidatorError[] = [];
+  const envVars = new Set();
+
+  const validate = (item: DbCreatorProps) => {
+    if (!/^[0-9a-z._-]{1,64}$/.test(item.name)) {
+      errors.push({
+        item,
+        message: `[${item.name}] is not a valid handle: /\A[0-9a-z._-]{1,64}\z/`,
+      });
+    }
+
+    if (envVars.has(item.env)) {
+      errors.push({
+        item,
+        message: `${item.env} has already been used, each database env var must be unique`,
+      });
+    } else {
+      envVars.add(item.env);
+    }
+  };
+
+  items.forEach(validate);
+  return errors;
+};
+
+const validateEnvs = (items: TextVal[]): ValidatorError[] => {
+  const errors: ValidatorError[] = [];
+
+  const validate = (item: TextVal) => {
+    // https://stackoverflow.com/a/2821201
+    if (!/[a-zA-Z_]+[a-zA-Z0-9_]*/.test(item.key)) {
+      errors.push({
+        item,
+        message: `${item.key} does not match regex: /[a-zA-Z_]+[a-zA-Z0-9_]*/`,
+      });
+    }
+
+    if (item.value === "") {
+      errors.push({
+        item,
+        message: `${item.key} is blank, either provide a value or remove the environment variable`,
+      });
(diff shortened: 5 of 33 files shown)


src/app/test/create-project.test.tsx
```diff
@@ -69,6 +89,8 @@ describe("Create project flow", () => {
       const dbEnvVar = await screen.findByRole("textbox", { name: "envvar" });
       expect(dbEnvVar).toHaveDisplayValue("DATABASE_URL");
 
+      await sleep(0);
```
madhuravius, line 92: 🤔

src/deploy/index.ts
```diff
@@ -20,3 +20,4 @@ export * from "./release";
 export * from "./container";
 export * from "./vpc_peer";
 export * from "./vpn_tunnel";
+export * from "./code-scan-result";
```
madhuravius, line 23: is there a way we don't have to do this? doesn't pruning automatically take care of this?

if this is stylistic we can probably write a utility that runs this on lint. i say this because i constantly forget this

src/routes/index.ts
```diff
@@ -135,6 +135,9 @@ export const environmentBackupsUrl = (id: string) =>
 export const ENVIRONMENT_SETTINGS_PATH = "/environments/:id/settings";
 export const environmentSettingsUrl = (id: string) =>
   `/environments/${id}/settings`;
+export const ENVIRONMENT_CREATE_APP_PATH = "/environments/:id/app";
+export const environmentCreateAppUrl = (id: string) =>
+  `/create/name?environment_id=${id}`;
```
madhuravius, line 140: why not make this a path parameter? is there ever a context we'll create an app outside of environment? if so i think it should be a path parameter and NOT a qsp (ftux?)

src/test/index.tsx
```diff
@@ -128,3 +128,13 @@ export const waitForToken = (store: Store<AppState>) => {
     }
   });
 };
+
+export const sleep = (n: number) =>
```
madhuravius, line 132: we should comment why this is needed, i'm a bit torn by it b/c i thought part of react testing library was that it waited for certain things to happen, why are we waiting like this? i'm sure there's a good reason i just dont' know why

src/ui/hooks/use-latest-code-results.ts
```diff
@@ -0,0 +1,21 @@
+import { fetchAppOperations, selectLatestSucceessScanOp } from "@app/deploy";
+import {
+  DeployCodeScanResponse,
+  fetchCodeScanResult,
+} from "@app/deploy/code-scan-result";
+import { AppState } from "@app/types";
+import { useSelector } from "react-redux";
+import { useCache, useQuery } from "saga-query/react";
+
+export const useLatestCodeResults = (appId: string) => {
```
madhuravius, line 10: 🤔 were you planning on adding hooks for the other fetches we do now? this gets you: a consolidated fetch + selector 

is the purpose of this to dry those where we do this (i do this for charts on apps/dbs for example), would you want hooks for those?

madhuravius: ex: `useMetricsByContainerIdForHorizon` or something

src/ui/pages/create-project-name.tsx
```diff
@@ -0,0 +1,281 @@
...
+    if (!validHandle(name)) {
+      setNameError(
+        "Not a valid app name (letters, numbers, and following symbols (., -, _).",
+      );
+      return;
+    } else {
+      setNameError("");
+    }
+
+    thunk.trigger();
```
madhuravius, line 61: TIL i was wondering how to do this

src/app/test/create-project.test.tsx
madhuravius: i think i might've used this or added it, but probably did not need to at the time (for logout?) we should probably see why we need this or explain somewhere i think

src/test/index.tsx
madhuravius: updated comment: https://github.com/aptible/app-ui/pull/222/files#r1270691778

src/ui/pages/create-project-name.tsx
```diff
@@ -0,0 +1,281 @@
...
+    selectStackById(s, { id: env.stackId }),
+  );
+  const [name, setName] = useState("");
+  const [nameError, setNameError] = useState("");
+  const thunk = useApi(createDeployApp({ name, envId }));
+  const onSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!validHandle(name)) {
+      setNameError(
```
madhuravius, line 53: we should probably move this to a util and make it accept a regex imo, we will use this everywhere

if not in this pr, one that immediately follows should employ this so we don't reimplement this everywhere, to the container (maybe with a ref? or something so we can set errors on it independent of child component)

src/ui/pages/create-project-name.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const [nameError, setNameError] = useState("");
+  const thunk = useApi(
+    createProject({ name, stackId: stackValue.value, orgId: org.id }),
+  );
+  const onSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!validHandle(name)) {
+      setNameError(
+        "Not a valid environment name (letters, numbers, and following symbols (., -, _).",
```
madhuravius, line 181: see same as above, i can see us rewriting this for every new component and would like to avoid repeating this kind of static content, but lgtm

src/ui/pages/create-project-name.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const defaultStack = useSelector(selectStackPublicDefault);
+  const stackId = queryStackId || defaultStack.id;
+
+  useQuery(fetchAllStacks());
+
+  if (queryEnvId === "") {
+    return <CreateEnvironmentPage stackId={stackId} />;
+  }
+
+  return <CreateAppPage envId={queryEnvId} />;
```
madhuravius, line 280: you probably want to wrap this in a LoadingResources wrapper, i forget its exact name, but waiting for the query for fetch all stacks to complete, right?

src/ui/shared/select.tsx
```diff
@@ -19,6 +20,7 @@ export function Select({
   onSelect,
   defaultValue,
   className = "",
+  ariaLabel = "combobox",
```
madhuravius, line 23: usually selects are form components, so i think we should probably just use the label? you don't need an arialabel if you have a labeled-for / labeled-by and inherit that label i think (ex: https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/) that way the label will work and you can label it with the label component you wrote before i think

madhuravius: i say that because the labels will drift and it will suck to go back and fix all the content

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+  Input,
+  Loading,
+  PreCode,
+  ProgressProject,
+  Select,
+  SelectOption,
+  tokens,
+} from "../shared";
+
+const trim = (t: string) => t.trim();
```
madhuravius, line 47: i'm confused by this, is this solely for the purpose of having an easier `.map()`? because normally I'd coerce it in `.map((str) => str.trim())` unless you have a strong reason for not doing that (doesn't really add LOC and is minimal in reading to me)

that being said, if you're going to do this and feel strongly, do the same with the others (filterFalsies) or something, i think consistency here is important

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+  message: string;
+  item: DbCreatorProps;
+}
+
+const validateDbs = (items: DbCreatorProps[]): DbValidatorError[] => {
+  const errors: DbValidatorError[] = [];
+  const envVars = new Set();
+
+  const validate = (item: DbCreatorProps) => {
+    if (!/^[0-9a-z._-]{1,64}$/.test(item.name)) {
```
madhuravius, line 86: we should make this regex passable and set this as a constant (HANDLE_VALIDATION) or something

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+      errors.push({
+        item,
+        message: `[${item.name}] is not a valid handle: /\A[0-9a-z._-]{1,64}\z/`,
```
madhuravius, line 89: i would reference the regex value that's a const here to avoid drift

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+        value={db.env}
+        onChange={(value) => propChange({ ...db, env: value })}
+      />
+    </div>
+  );
+
+  const imgOptions = [
+    { value: "", label: "Choose a Database" },
+    ...images.map((img) => ({
+      label: `${img.type} v${img.version}`,
```
madhuravius, line 285: img could be  a falsy / undefined type, so you might wanan do the same thing you see on line 265 or it would say `undefined`? (unless our types are wrong, right?)

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+  const [searchParams] = useSearchParams();
+  const queryDbsStr = searchParams.get("dbs") || "";
+  // prefill databases based on query params
+  useEffect(() => {
+    if (!queryDbsStr) return;
+    if (!namePrefix) return;
+    const qdbs = queryDbsStr.split(",");
+    if (qdbs.length === 0) return;
+
+    // dbDispatch({ type: "reset" });
```
madhuravius, line 431: desired or maybe for testing at the time?

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+        payload: {
+          env: env.toLocaleUpperCase(),
+          id: `${createId()}`,
+          imgId: img.id,
+          name: `${namePrefix}-${img.type || ""}`,
+          dbType: img.type || "",
+        },
+      });
+    });
+  }, [queryDbsStr, dbImages, namePrefix]);
```
madhuravius, line 449: this is an array of objects and if they ever mutate or hydrate this will fire again, so you might wanna serialize / generate a common string or a memoization technique, right?

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+          >
+            <div className="flex flex-col gap-4">
+              <Loading
+                text="Loading databases..."
+                isLoading={dbsQuery.isInitialLoading}
+              />
+              {existingDbs.length > 0 ? (
+                <DatabaseExistingForm
+                  envId={app.environmentId}
+                  dbMap={dbExistingMap}
```
madhuravius, line 857: i was having a hard time tracking the existing maps vs creator maps, are they actually different? are they the same? if they're different having different names might make this less confusing

src/ui/pages/create-project-git-settings.tsx
```diff
@@ -0,0 +1,952 @@
...
+        variant="secondary"
+        isLoading={isLoading}
+      >
+        <IconPlusCircle className="mr-2" color="#fff" /> New Database
+      </Button>
+    </div>
+  );
+};
+
+const DockerfileDataView = ({
```
madhuravius, line 495: todo (not necessarily in this pr), but we should move this + the hook to another place so we can reference this in other views as i can see it being useful for viewing what an app is running at a glance (maybe settings?)

src/ui/pages/create-project-name.tsx
```diff
@@ -0,0 +1,281 @@
+import { useEffect, useState } from "react";
```
madhuravius, line 1: component name is a bit weird, i'd probably call it create-app.tsx or something

madhuravius (commented):
minor reminder, there's some new items in here, might be worthwhile updating the styles.tsx (i reference it more and more as i look for things as i'm building, i've noticed, or using it as a discussion focus w/ joe)

src/app/test/create-project.test.tsx
neurosnap: done

src/deploy/index.ts
neurosnap: It's about creating module boundaries and not reaching into specific files.  Always import from `index` for modules.

src/routes/index.ts
neurosnap: `/create/name` is already a route, we just wanted to add a query param.  This is also a nice name to reuse in other places.

src/ui/hooks/use-latest-code-results.ts
neurosnap: You're welcome to wrap queries and selectors inside hooks.

src/ui/pages/create-project-git-settings.tsx
neurosnap: removed

src/ui/pages/create-project-git-settings.tsx
neurosnap: are you referring to `dbImages`?  That should already be memoized with `createSelector`

src/ui/pages/create-project-git-settings.tsx
neurosnap: I tried a couple ways to accomplish this, I landed on this one because it is less jank -- but kind of annoying.

src/ui/pages/create-project-name.tsx
neurosnap: I'm not a fan of having loaders block views.  The views will react properly when the stacks are loaded.

src/test/index.tsx
neurosnap: I needed it to hack some tests but they ended up being no necessary.  I'd like to leave this here to so we can use it to debug.

src/routes/index.ts
madhuravius: understood, i can see there being a case for us to do this, i just didn't realize you were thinking about that here or if your ideas were tainted by my middleware/qsp request to you the other day

src/ui/hooks/use-latest-code-results.ts
madhuravius: i will have to do this for metrics because of the number of things it hits, may need your help in reviewing that cleanup

src/ui/pages/create-project-name.tsx
madhuravius: 🤔 understood, i haven't been a fan of the react shuffle dance that tends to happen with this but maybe our api/backend is so fast that we don't notice this

madhuravius (approved):
LGTM - tested locally and ensured there was minimal pain while testing

we will have a pt2/3 probably of cleanup and more refactor

PR #228, opened 2023-07-23
Diy chart

neurosnap:
Screenshot 2023-07-24 at 10 14 59 AM

### DIFF
--- a/src/ui/shared/container-metrics-chart.tsx
+++ b/src/ui/shared/container-metrics-chart.tsx
@@ -1,5 +1,5 @@
+import cn from "classnames";
 import { Line } from "react-chartjs-2";
-import cn from 'classnames';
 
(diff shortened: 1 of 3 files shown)


src/ui/shared/container-metrics-chart.tsx
```diff
@@ -29,7 +31,200 @@ ChartJS.register(
...
+  if (datasets.length === 0) {
+    return <div>no data</div>;
+  }
+
+  // options
+  const tickMax = 5;
+  const margin = { top: 30, bottom: 30, left: 30, right: 30 };
+  const strokeWidth = 2;
+
+  const h = 400;
```
madhuravius, line 71: i think we should have a oop-ey flavor of this code because it's kind of hard to keep track of all the arithmetic here, i'd stick this in an object (plotconfig) or something and have it computed and referenced from there (and unit tested)

PR #231, opened 2023-07-25
feat: manage endpoints

neurosnap:
https://app.shortcut.com/aptible/story/14116/manage-endpoints

### DIFF
--- a/src/deploy/activity/index.ts
+++ b/src/deploy/activity/index.ts
@@ -2,6 +2,7 @@ import { createSelector } from "@reduxjs/toolkit";
 
 import { findAppById, selectApps } from "../app";
 import { findDatabaseById, selectDatabases } from "../database";
+import { findEndpointById, getEndpointUrl, selectEndpoints } from "../endpoint";
 import {
   findEnvById,
   hasDeployEnvironment,
@@ -121,7 +122,8 @@ export const selectResourceNameByOperationId = createSelector(
   selectOperationById,
   selectDatabases,
   selectApps,
-  (op, dbs, apps) => {
+  selectEndpoints,
+  (op, dbs, apps, enps) => {
     let resourceHandle = "";
     if (op.resourceType === "app") {
       const app = findAppById(apps, { id: op.resourceId });
@@ -129,6 +131,9 @@ export const selectResourceNameByOperationId = createSelector(
     } else if (op.resourceType === "database") {
       const db = findDatabaseById(dbs, { id: op.resourceId });
       resourceHandle = db.handle;
+    } else if (op.resourceType === "vhost") {
+      const enp = findEndpointById(enps, { id: op.resourceId });
+      resourceHandle = getEndpointUrl(enp);
     } else {
       resourceHandle = op.resourceId;
     }
--- a/src/deploy/endpoint/index.ts
+++ b/src/deploy/endpoint/index.ts
@@ -1,48 +1,58 @@
+import { createAction, createSelector } from "@reduxjs/toolkit";
+
+import { api, cacheShortTimer, thunks } from "@app/api";
 import {
   call,
+  poll,
   put,
   setLoaderError,
   setLoaderStart,
   setLoaderSuccess,
 } from "@app/fx";
-
-import { api, cacheShortTimer, thunks } from "@app/api";
-import { defaultEntity, extractIdFromLink } from "@app/hal";
+import { defaultEntity, defaultHalHref, extractIdFromLink } from "@app/hal";
 import {
   createReducerMap,
   createTable,
   mustSelectEntity,
 } from "@app/slice-helpers";
 import type {
+  AcmeConfiguration,
+  AcmeStatus,
   AppState,
   DeployEndpoint,
   DeployOperationResponse,
   LinkResponse,
   ProvisionableStatus,
 } from "@app/types";
-import { createSelector } from "@reduxjs/toolkit";
 
 import { selectAppById, selectAppsByEnvId } from "../app";
+import { createCertificate } from "../certificate";
 import { selectDatabasesByEnvId } from "../database";
 import { selectDeploy } from "../slice";
 
 interface DeployEndpointResponse {
   id: number;
   acme: boolean;
-  acme_configuration: string;
+  acme_configuration: AcmeConfiguration | null;
   acme_dns_challenge_host: string;
   acme_status: string;
-  container_exposed_ports: string[];
-  container_port: string;
+  container_exposed_ports: string[] | null;
+  container_port: string | null;
   container_ports: string[];
   default: boolean;
-  docker_name: string;
+  docker_name: string | null;
+  elastic_load_balancer_name: string | null;
   external_host: string;
-  external_http_port: string;
-  external_https_port: string;
+  external_http_port: string | null;
+  external_https_port: string | null;
   internal: boolean;
+  internal_health_port: string | null;
+  internal_host: string | null;
+  internal_http_port: string | null;
+  internal_https_port: string | null;
   ip_whitelist: string[];
   platform: "alb" | "elb";
+  security_group_id: string;
   type: string;
   user_domain: string;
   virtual_domain: string;
@@ -63,7 +73,7 @@ export const defaultEndpointResponse = (
   return {
     id: 0,
     acme: false,
-    acme_configuration: "",
+    acme_configuration: null,
     acme_status: "",
     acme_dns_challenge_host: "",
     container_exposed_ports: [],
@@ -71,21 +81,27 @@ export const defaultEndpointResponse = (
     container_ports: [],
     default: true,
     docker_name: "",
+    elastic_load_balancer_name: "",
     external_host: "",
     external_http_port: "",
     external_https_port: "",
     internal: false,
+    internal_health_port: "",
+    internal_host: "",
+    internal_http_port: "",
+    internal_https_port: "",
     ip_whitelist: [],
     platform: "elb",
     type: "",
     user_domain: "",
     virtual_domain: "",
+    security_group_id: "",
     status: "unknown",
     created_at: now,
     updated_at: now,
     _links: {
-      service: { href: "" },
-      certificate: { href: "" },
+      service: defaultHalHref(),
+      certificate: defaultHalHref(),
     },
     _type: "vhost",
     ...resp,
@@ -100,16 +116,16 @@ export const deserializeDeployEndpoint = (
     acme: payload.acme,
     acmeConfiguration: payload.acme_configuration,
     acmeDnsChallengeHost: payload.acme_dns_challenge_host,
-    acmeStatus: payload.acme_status,
+    acmeStatus: payload.acme_status as AcmeStatus,
     containerExposedPorts: payload.container_exposed_ports,
-    containerPort: payload.container_port,
+    containerPort: payload.container_port || "",
     containerPorts: payload.container_ports,
     createdAt: payload.created_at,
     default: payload.default,
-    dockerName: payload.docker_name,
+    dockerName: payload.docker_name || "",
     externalHost: payload.external_host,
-    externalHttpPort: payload.external_http_port,
-    externalHttpsPort: payload.external_https_port,
+    externalHttpPort: payload.external_http_port || "",
+    externalHttpsPort: payload.external_https_port || "",
     internal: payload.internal,
     ipWhitelist: payload.ip_whitelist,
     platform: payload.platform,
@@ -131,9 +147,9 @@ export const defaultDeployEndpoint = (
     id: "",
     status: "pending",
     acme: false,
-    acmeConfiguration: {},
+    acmeConfiguration: null,
     acmeDnsChallengeHost: "",
-    acmeStatus: "",
+    acmeStatus: "pending",
     containerExposedPorts: [],
     containerPort: "",
     containerPorts: [],
@@ -167,8 +183,13 @@ const selectors = slice.getSelectors(
 );
 const initApp = defaultDeployEndpoint();
 const must = mustSelectEntity(initApp);
+const {
+  selectTable: selectEndpoints,
+  selectTableAsList: selectEndpointsAsList,
+} = selectors;
 export const selectEndpointById = must(selectors.selectById);
-export const { selectTableAsList: selectEndpointsAsList } = selectors;
+export const findEndpointById = must(selectors.findById);
+export { selectEndpoints, selectEndpointsAsList };
 export const hasDeployEndpoint = (a: DeployEndpoint) => a.id !== "";
 export const endpointReducers = createReducerMap(slice);
 
@@ -227,6 +248,25 @@ export const selectEndpointByEnvironmentAndCertificateId = createSelector(
     endpoints.filter((endpoint) => endpoint.certificateId === certificateId),
 );
 
+export const selectAppsByCertificateId = createSelector(
+  selectAppsByEnvId,
+  selectEndpointsByEnvironmentId,
+  (_: AppState, p: { certificateId: string }) => p.certificateId,
+  (apps, endpoints, certificateId) => {
+    const endpointsWithCertificates = endpoints.filter(
+      (endpoint) => endpoint.certificateId === certificateId,
+    );
+
+    return apps.filter((app) => {
+      return app.serviceIds.some((appServiceId) => {
+        return endpointsWithCertificates.find(
+          (endpoint) => endpoint.serviceId === appServiceId,
+        );
+      });
+    });
+  },
+);
+
 export const fetchEndpointsByAppId = api.get<{ appId: string }>(
   "/apps/:appId/vhosts",
   { saga: cacheShortTimer() },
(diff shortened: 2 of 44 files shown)


src/deploy/activity/index.ts
```diff
@@ -121,14 +122,18 @@ export const selectResourceNameByOperationId = createSelector(
   selectOperationById,
   selectDatabases,
   selectApps,
-  (op, dbs, apps) => {
+  selectEndpoints,
+  (op, dbs, apps, enps) => {
```
madhuravius, line 126: *chef's kiss here*, so this means we'll get the endpoint operations on this list that was requested it seems

src/deploy/endpoint/index.ts
```diff
@@ -227,6 +248,25 @@ export const selectEndpointByEnvironmentAndCertificateId = createSelector(
...
+export const selectAppsByCertificateId = createSelector(
+  selectAppsByEnvId,
+  selectEndpointsByEnvironmentId,
+  (_: AppState, p: { certificateId: string }) => p.certificateId,
+  (apps, endpoints, certificateId) => {
+    const endpointsWithCertificates = endpoints.filter(
+      (endpoint) => endpoint.certificateId === certificateId,
+    );
+
+    return apps.filter((app) => {
```
madhuravius, line 260: we should rewrite this in the format you like at some point, reading this i'm still not quite sure how it works (you and i paired through it and both of us were sort of scratching our heads as to how this ends up working)

src/deploy/endpoint/index.ts
```diff
@@ -254,25 +301,62 @@ export const endpointEntities = {
...
+export type CreateEndpointProps =
+  | CreateDefaultEndpoint
+  | CreateManagedEndpoint
+  | CreateCustomEndpoint;
+
+export const createEndpoint = api.post<
+  CreateEndpointProps & { certId: string }
+>("/services/:serviceId/vhosts", function* (ctx, next) {
+  const data = {
+    platform: "alb",
```
madhuravius, line 340: this is specifically http / http_proxy_protocol endpoints right? i think we have to use elb for tcp  (i think?) or it will error

we may not be here yet, but i think if we offer create, we should allow tcp endpoints too, right (unless we do not offer this right now)

src/deploy/endpoint/index.ts
```diff
@@ -348,3 +469,191 @@ export const provisionEndpoint = thunks.create<CreateEndpointProps>(
...
+    });
+    ctx.request = ctx.req({ body });
+
+    yield* next();
+  },
+);
+
+/**
+ *
+ * database vhost
```
madhuravius, line 510: nit - reminder to delete or set type/test and move on, just clean up

src/deploy/endpoint/index.ts
```diff
@@ -348,3 +469,191 @@ export const provisionEndpoint = thunks.create<CreateEndpointProps>(
...
+    ctx.loader = { id: ctx.key, meta: { opId: opCtx.json.data.id } };
+    yield* next();
+  },
+);
+
+export function requiresAcmeSetup(enp: DeployEndpoint) {
+  return enp.acme && enp.acmeStatus !== "ready" && enp.status === "provisioned";
+}
+
+function normalizeName(name: string) {
```
madhuravius, line 568: nitpick - confusing function name, not sure what it's really meant to do (i would call it ensureTrailingPeriod or something?)

src/deploy/endpoint/index.ts
```diff
@@ -348,3 +469,191 @@ export const provisionEndpoint = thunks.create<CreateEndpointProps>(
...
+interface DnsAnswer {
+  data: string;
+}
+
+export const checkDns = thunks.create<{ from: string; to: string }>(
+  "check-dns",
+  function* (ctx, next) {
+    const { from, to } = ctx.payload;
+    yield* put(setLoaderStart({ id: ctx.key }));
+    const rand = Math.floor(Math.random() * 10000);
```
madhuravius, line 584: 🤔 a good reason for this? might wanna leave a note

src/deploy/endpoint/index.ts
```diff
@@ -348,3 +469,191 @@ export const provisionEndpoint = thunks.create<CreateEndpointProps>(
...
+export const getPlacement = (enp: DeployEndpoint) => {
+  if (enp.externalHost) {
+    return "External (publicly accessible)";
+  }
+
+  return "Internal";
+};
+
+export const getIpAllowlistText = (enp: DeployEndpoint) => {
+  return enp.ipWhitelist.length ? enp.ipWhitelist.join(", ") : "Disabled";
```
madhuravius, line 619: stylistic nit - in other places ive noticed you like to do length === 0

src/deploy/endpoint/index.ts
```diff
@@ -348,3 +469,191 @@ export const provisionEndpoint = thunks.create<CreateEndpointProps>(
...
+};
+
+export const getContainerPort = (
+  enp: DeployEndpoint,
+  exposedPorts: string[],
+) => {
+  let port = "Unknown";
+  if (exposedPorts.length > 0) {
```
madhuravius, line 627: i found out today we do not choose the first port, but the lowest port ordered sequentially, not sure if you want to honor this

src/ui/layouts/app-detail-layout.tsx
```diff
@@ -64,7 +64,7 @@ export function AppHeader({ app }: { app: DeployApp }) {
             : "Unknown"}
         </DetailInfoItem>
         <DetailInfoItem title="Docker Image">
-          {app.currentImage?.dockerRepo}
+          {app.currentImage.dockerRepo}
```
madhuravius, line 67: sanity check/reminder for me to verify in staging - you may want to verify this is a not null value on backend, this was probably me when we were testing FTUX and i went back to view apps that failed to deploy on their initial deploy (people can get in a weird state) at which point you might have a bad time on this page

src/ui/pages/app-create-endpoint.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const [port, setPort] = useState("");
+  const [enpType, setEnpType] = useState<EndpointType>("default");
+  const [enpPlacement, setEnpPlacement] = useState("external");
+  const [ipAllowlist, setIpAllowlist] = useState("");
+  const [domain, setDomain] = useState("");
+  const [transCert, setTransCert] = useState(false);
+  const [cert, setCert] = useState("");
+  const [privKey, setPrivKey] = useState("");
+
+  const createData = (): CreateEndpointProps => {
```
madhuravius, line 63: cleanup/nit - i would put this in the deploy package with those parameters supplied in an object (as it is returning props specifically from that file)

src/ui/hooks/use-validator.ts
```diff
@@ -0,0 +1,86 @@
...
+ *   },
+ *  };
+ *
+ *  function App() {
+ *   const [domain, setDomain] = useState("");
+ *   const [errors, validate] = useValidator<FormData, typeof validators>(validators);
+ *   const data = { domain };
+ *
+ *   const onSubmit = () => {
+ *     if (!validate(data)) return;
```
madhuravius, line 31: 👌 great, lgtm, easy enough to pattern match and make a ton of these properly setup

i think it's not a bad idea to start having some recyclable validator utils so it's just easy to group them together in an array and chain them in an array of validations before this file calls that. can chat with you about it later, but this is exactly what i hoped you would do

src/ui/pages/app-create-endpoint.tsx
```diff
@@ -0,0 +1,272 @@
...
+      return {
+        type: "managed",
+        cert,
+        privKey,
+        domain,
+        ...def,
+      };
+    }
+
+    if (enpType === "custom") {
```
madhuravius, line 81: is there a reason spread operator order is different here? neither of them seem to have overlapping keys with `def`

src/ui/pages/create-project-git.tsx
```diff
@@ -786,27 +785,14 @@ const CreateEndpointForm = ({ app }: { app: DeployApp }) => {
...
-            </label>
-          </div>
-        );
-      })}
+      <CreateAppEndpointSelector
+        app={app}
+        selectedId={curServiceId}
+        onSelect={(id: string) => setServiceId(id)}
+        disabled={(service) =>
+          !!vhosts.find((vhost) => vhost.serviceId === service.id)
```
madhuravius, line 793: nit - isn't this just the `.some`?

src/ui/shared/endpoint/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    "text-sm font-semibold ",
+    "px-2 flex justify-between items-center w-fit",
+  );
+
+  if (
+    status === "provisioning" ||
+    status === "deprovisioning" ||
+    status === "pending"
+  ) {
+    return (
```
madhuravius, line 31: sanity check - not sure if you want to use it but there was a Pill component i moved into shared that was prebaked with some colors (might be easier to use than this), but not sure

src/ui/shared/endpoint/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  if (enp.status === "provisioning") {
+    return <p className="text-gray-500 italic">Provisioning</p>;
+  }
+
+  if (enp.type === "tcp") {
+    return <span>{enp.externalHost}</span>;
+  }
+
+  return (
+    <ExternalLink href={`https://${enp.virtualDomain}`} variant="default">
```
madhuravius, line 77: ~reminder to check - i don't think all endpoints are https, some are tcp flavors (not sure, need to check). i'm not sure if prefix by protocol (`postgresql://` or `redis://`/`rediss://` etc)~

~i know we don't support udp~

madhuravius: nvm you address this here: https://github.com/aptible/app-ui/pull/231/files#diff-d631e3f1bc2361ac2fe24d5d912001c193290769e4ff4482da200dace4291d24R30

src/deploy/activity/index.ts
```diff
@@ -2,6 +2,7 @@ import { createSelector } from "@reduxjs/toolkit";
 
 import { findAppById, selectApps } from "../app";
 import { findDatabaseById, selectDatabases } from "../database";
+import { findEndpointById, getEndpointUrl, selectEndpoints } from "../endpoint";
```
madhuravius, line 5: for later, i can't easily get to the activity page when trying out this pr:

image

src/ui/pages/endpoint-detail-settings.tsx
```diff
@@ -0,0 +1,93 @@
...
+  };
+  useLoaderSuccess(loader, () => {
+    navigate(endpointDetailActivityUrl(id));
+  });
+
+  return (
+    <div className="flex flex-col gap-4">
+      <form onSubmit={onSave}>
+        <FormGroup label="Container Port" htmlFor="port">
+          <Input
```
madhuravius, line 57: nit - i think this one needs validation (the field is not a number field and it's not really 1-65535 or whatever)

image

src/ui/pages/endpoint-detail-settings.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </FormGroup>
+        <ButtonCreate
+          type="submit"
+          envId={service.environmentId}
+          isLoading={loader.isLoading}
+        >
+          Save
+        </ButtonCreate>
+
+        <BannerMessages {...loader} />
```
madhuravius, line 82: nit - need spacing if this is around

image

src/ui/shared/endpoint/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const loader = useLoader(action);
+  const onClick = () => {
+    dispatch(action);
+  };
+  useLoaderSuccess(loader, () => {
+    navigate(operationDetailUrl(loader.meta.opId));
+  });
+
+  return (
+    <ButtonDestroy
```
madhuravius, line 99: you really should force a confirm on this one i think, it just GOES:

image

image

if deprovisioning, should be disabled

src/ui/shared/endpoint/index.tsx
madhuravius: this link is confusing (an ID href with no indication that's where you'd go for details, right next to a link that'll take you to the endpoint), something to revisit w/ joe unless you have a simple solution

image

madhuravius: it may also be confusing because we don't use this link id pattern in many places (?) not sure

either way this doesn't have to be actioned now, just something to return to

src/ui/shared/endpoints-overview.tsx
```diff
@@ -1,166 +1,112 @@
...
-      <div className="flex mb-4">
-        {endpoints.length ? (
-          <InputSearch
-            className="self-end"
-            placeholder="Search endpoints..."
-            search={search}
-            onChange={onChange}
-          />
-        ) : null}
+      <div className="flex justify-between items-center mb-4">
```
madhuravius, line 103: this wasn't working for me:

image

src/ui/shared/activity.tsx
```diff
@@ -389,3 +391,37 @@ export function ActivityByDatabase({ dbId }: { dbId: string }) {
     </LoadResources>
   );
 }
+
+export function ActivityByEndpoint({ enpId }: { enpId: string }) {
```
madhuravius, line 395: people will raise this most likely, but i'm sure you're aware it doesn't appear in the global search results (but there's no indication on that page that it excludes these ops/activity), so you may want to return to this:

image

src/ui/pages/endpoint-detail-settings.tsx
```diff
@@ -0,0 +1,93 @@
...
+            name="ip-allowlist"
+            value={ipAllowlist}
+            onChange={(e) => setIpAllowlist(e.currentTarget.value)}
+          />
+        </FormGroup>
+        <ButtonCreate
+          type="submit"
+          envId={service.environmentId}
+          isLoading={loader.isLoading}
+        >
```
madhuravius, line 78: upon save i can't edit again:

image

with no error

madhuravius: was forced to reload (i think empty container port - not defaulting is causing issues, as an empty field is allowed)

src/ui/shared/endpoint/index.tsx
madhuravius: i also dont think you got it to work on all cases:

image

I have pending operations on this one and it's not showing

image

src/ui/pages/app-create-endpoint.tsx
```diff
@@ -0,0 +1,272 @@
...
+    if (data.cert === "") return "A certificate is required for custom HTTPS";
+  },
+  privKey: (data: CreateEndpointProps) => {
+    if (data.type !== "custom") return;
+    if (data.privKey === "")
+      return "A private key is required for custom HTTPS";
+  },
+};
+
+export const AppCreateEndpointPage = () => {
```
madhuravius, line 46: reminder - i see app endpoints, but not db create flows (it's missing though it's on the ticket):

image

i can't get to create it from here  (looks like there's no page?)

src/ui/shared/endpoints-overview.tsx
madhuravius: btw for consistency we don't allow search on this type of page on all other resources' children:

image

madhuravius (changes requested):
LGTM 

i found a few bugs (those are the only things i think are worthy of fixing)

everything else (nit/style) can be done in a followup

also there's no db endpoints (though the ticket seems to ask for it) - may wanna make a followup for that OR make another ticket

src/deploy/endpoint/index.ts
neurosnap: Once we get to databases we need `tcp` and `elb`.  Is it ok if we save for then?

src/ui/layouts/app-detail-layout.tsx
neurosnap: We now guarantee the object exists in the UI.

src/ui/hooks/use-validator.ts
neurosnap: Yep, this will allow us to do that.

src/ui/pages/app-create-endpoint.tsx
neurosnap: no reason, updating.

src/ui/shared/endpoint/index.tsx
neurosnap: The status is on the vhost not the operation.  We would have to do some polling to keep that table up-to-date.

src/ui/shared/endpoints-overview.tsx
neurosnap: removed

src/deploy/endpoint/index.ts
madhuravius: yeah part 2 is fine just don't forget imo

src/ui/shared/endpoint/index.tsx
madhuravius: no i was reloading the page and it was not appearing, one of the statuses is missing in your switch statement (i think)

src/deploy/activity/index.ts
madhuravius: it's more for later in terms of how users can discover / work with stuff on the app as opposed to knowing things exist to do things with

src/ui/shared/endpoint/index.tsx
madhuravius: i like it myself but we should click it twice IMO (Deprovision + Confirm)

src/ui/pages/app-create-endpoint.tsx
madhuravius: looks like a part 2 or 3 for this one

src/ui/pages/endpoint-detail-settings.tsx
madhuravius: in discussion looks like this was resolved

src/ui/pages/endpoint-detail-settings.tsx
madhuravius: also was apparently resolved

madhuravius (approved):

PR #238, opened 2023-07-31
feat: dedicated create db page

neurosnap:


joeclayallday:
This looks dope! Nice work @neurosnap 

I'll be back online this Wednesday. Just wanted to share notes so you can keep moving fast:
• Add a cancel button to the right of the save button in our white secondary button style, so users can exit without saving
• After clicking "New Database" show the form already open so users don't have to click the blue "New Database" button to see choose their database (saves a click)
• I'd wrap the whole form in a div with these classes: `bg-white py-8 px-8 shadow border border-black-100 rounded-lg`

src/ui/shared/db/creator-form.tsx
```diff
@@ -0,0 +1,264 @@
...
+            />
+          );
+        })}
+      <Button
+        type="button"
+        onClick={onClick}
+        variant="secondary"
+        isLoading={isLoading}
+      >
+        <IconPlusCircle className="mr-2" color="#fff" variant="sm" /> New
```
madhuravius, line 259: form validation is a bit off here: 

image

madhuravius (commented):
LGTM - stylistic changes

src/deploy/database/index.ts
```diff
@@ -387,6 +395,38 @@ interface CreateDbResult {
...
+    databaseImageId: db.imgId,
+  };
+};
+
+export const provisionDatabaseList = thunks.create<{
+  envId: string;
+  dbs: DbCreatorProps[];
+}>("database-list-provision", function* (ctx, next) {
+  const { dbs, envId } = ctx.payload;
+  yield* put(setLoaderStart({ id: ctx.key }));
```
madhuravius, line 417: there is a race condition that seems to be possible. I created 3 in the list at once, and only 2 were sent to operations queue

neurosnap: We were swallowing the errors so nice catch!  I've updated the code to send all errors detected.

madhuravius: when this errors i still don't see the errors btw

madhuravius: ex ( have one with this name already, i see no error ): 

image

neurosnap: I specifically tested this scenario, are you sure you pulled latest?

src/ui/shared/db/creator-form.tsx
neurosnap: I don't understand what I'm looking at here.  Why does it have a bunch of characters instead of the banner?

src/deploy/database/index.ts
neurosnap: Oh, you need to do a `yarn install` because I updated `saga-query`

madhuravius: i did but didn't restart the server, possible local caching issues

src/ui/shared/db/creator-form.tsx
madhuravius: resolved on pair

madhuravius (approved):

madhuravius (approved):

PR #241, opened 2023-08-02
feat: read and edit backup retention policy

neurosnap:
Screenshot 2023-08-02 at 3 39 04 PM

### DIFF
--- a/src/ui/shared/db/backup-list.tsx
+++ b/src/ui/shared/db/backup-list.tsx
@@ -1,19 +1,20 @@
-import { ReactElement } from "react";
-
-import {
-  InputSearch,
-  LoadResources,
-  ResourceListView,
-  TableHead,
-  Td,
-  tokens,
-} from "../";
-import { prettyEnglishDateWithTime } from "@app/date";
-import { BackupResponse } from "@app/deploy";
-import { LoadingState } from "@app/fx";
-import { capitalize } from "@app/string-utils";
 import cn from "classnames";
-import { useSearchParams } from "react-router-dom";
+import { ReactElement } from "react";
+import { useDispatch, useSelector } from "react-redux";
+import { Link } from "react-router-dom";
+import { useLoader } from "saga-query/react";
+
+import { prettyEnglishDateWithTime } from "@app/date";
+import { deleteBackup, selectDatabaseById } from "@app/deploy";
+import { databaseDetailUrl, operationDetailUrl } from "@app/routes";
+import { capitalize } from "@app/string-utils";
+import { AppState, DeployBackup } from "@app/types";
+
+import { BannerMessages } from "../banner";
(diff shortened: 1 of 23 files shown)


src/ui/shared/environment/backup-retention-policy-view.tsx
```diff
@@ -0,0 +1,170 @@
...
+      <h3 className={tokens.type.h3}>Backup Retention Policy</h3>
+
+      <form onSubmit={onSubmit} className="flex flex-col gap-2 mt-4">
+        <FormGroup
+          label="Daily backups retained"
+          htmlFor="daily"
+          feedbackMessage={errors.daily}
+          feedbackVariant={errors.daily ? "danger" : "info"}
+        >
+          <Input
```
madhuravius, line 104: since the minimum value is 0 somehow i cannot literally enter the number 2 in the form (i have to hit the up arrow or click the up arrow button)

we need to allow users to set a normal integer without blocking them probably

src/ui/shared/db/backup-list.tsx
```diff
@@ -36,107 +37,126 @@ const BackupTypePill = ({ manual }: { manual: boolean }): ReactElement => {
...
+const DeleteBackup = ({ envId, id }: { envId: string; id: string }) => {
+  const dispatch = useDispatch();
+  const action = deleteBackup({ id });
+  const loader = useLoader(action);
+  const onClick = () => {
+    dispatch(action);
+  };
+
+  return (
+    <ButtonDestroy
```
madhuravius, line 49: nit - the alert for this is super weird

image

should probably do something like tell you a delete operation is queued

madhuravius (approved):
lgtm, left nits / stylistic changes that can be done in a followup if desired, functions mostly as expected

joeclayallday:
@madhuravius @neurosnap 

We could add a button on each Database / Backups Tab that links back to the parent environment / backup management tab

Screen Shot 2023-08-03 at 6 07 09 PM

joeclayallday:
Can we add the total count of backups shown here above the table?

ex) 6 Backups

src/ui/shared/environment/backup-retention-policy-view.tsx
neurosnap: Hmm I'm able to do is in Chrome, are you using a different browser?

https://github.com/aptible/app-ui/assets/1940365/1ebc71b9-cde6-4281-b7be-bffc316c7b3d

madhuravius: i was discussing without doing the selection mechanic (tabbing to the field) or clicking on the edge of it

we resolved over pair, and fixed

madhuravius (approved):

PR #243, opened 2023-08-02
Add Security Dashboard link to Nav

joeclayallday:
This PR adds a sidebar nav link that points to `https://dashboard.aptible.com/controls` in deploy-ui. Users must log into deploy-ui to see the page.

Screen Shot 2023-08-02 at 3 38 44 PM


src/ui/shared/application-sidebar.tsx
```diff
@@ -44,6 +46,7 @@ export const ApplicationSidebar = () => {
     { name: "Databases", to: databaseUrl(), icon: <IconCylinder /> },
     { name: "Deployments", to: deploymentsUrl(), icon: <IconCloud /> },
     { name: "Activity", to: activityUrl(), icon: <IconHeart /> },
+    { name: "Security & Compliance", to: securityDashboardUrl(), icon: <IconShield /> },
```
neurosnap, line 49: I think we want to link directly to dashboard https://dashboard.aptible.com/controls

src/routes/index.ts
```diff
@@ -226,3 +226,6 @@ export const stackDetailVpcPeeringsUrl = (id: string) =>
 
 export const SEARCH_PATH = "/search";
 export const searchUrl = () => SEARCH_PATH;
+
+export const SECURITY_DASHBOARD_PATH = "/search";
+export const securityDashboardUrl = () => SECURITY_DASHBOARD_PATH;
```
neurosnap, line 231: I don't think we'll need these as we want to link directly to deploy-ui

neurosnap (changes requested):

src/routes/index.ts
joeclayallday: Thanks!

src/ui/shared/application-sidebar.tsx
joeclayallday: Thanks!

src/routes/index.ts
```diff
@@ -226,3 +226,6 @@ export const stackDetailVpcPeeringsUrl = (id: string) =>
 
 export const SEARCH_PATH = "/search";
 export const searchUrl = () => SEARCH_PATH;
+
+export const securityDashboardUrl = () =>
+  "https://dashboard.aptible.com/controls";
```
madhuravius, line 231: not sure if you want to do this but there's a slightly different value by env and we could make it easier for staging envs to go to their appropriate link (not that it matters too much for now)

madhuravius (commented):
lgtm, i think we could update that link for staging though if we wanted to (probably not worth the effort as we don't really establish staging security dashboards in most staging envs, right @neurosnap ?)

neurosnap (approved):

madhuravius (approved):

PR #252, opened 2023-08-07
Table Tweaks

joeclayallday:
Allow long TDs to wrap in our tables

BEFORE - TD content overflows
Screen Shot 2023-08-07 at 4 39 36 PM

AFTER
Screen Shot 2023-08-07 at 4 39 45 PM


madhuravius (approved):
LGTM - might wanna verify on the styles page as well

PR #253, opened 2023-08-08
fix: correct scale pages to actually work (missing tests/validations)

madhuravius:
* success banner poll (?) this could be really bad for something with a lot of services so i dropped those changes (rabbit hole)
* polling on app/db page for services (should do this in pair)
* validations (part 2?)

Mr sc 14408


src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -0,0 +1,321 @@
...
+  const app = useSelector((s: AppState) => selectAppById(s, { id }));
+  useQuery(fetchService({ id: serviceId }));
+  const service = useSelector((s: AppState) =>
+    selectServiceById(s, { id: serviceId }),
+  );
+
+  const onSubmitForm = (e: SyntheticEvent) => {
+    e.preventDefault();
+    if (!validate({ containerCount })) return;
+    dispatch(
```
madhuravius, line 58: TODO - need to properly present these errors

src/ui/pages/db-detail-scale.tsx
```diff
@@ -23,14 +55,34 @@ export const DatabaseScalePage = () => {
 
   const onSubmitForm = (e: SyntheticEvent) => {
     e.preventDefault();
+    if (!validate({ diskSize: diskValue })) return;
+    dispatch(
```
madhuravius, line 59: TODO need to properly present these errors

src/deploy/activity/index.ts
```diff
@@ -54,9 +65,9 @@ export const selectActivityForTableSearch = createSelector(
   selectActivityForTable,
   (_: AppState, props: { search: string }) => props.search.toLocaleLowerCase(),
   (_: AppState, props: { envId?: string }) => props.envId || "",
-  (_: AppState, props: { resourceId?: string }) => props.resourceId || "",
-  (ops, search, envId, resourceId): DeployActivityRow[] => {
-    if (search === "" && envId === "" && resourceId === "") {
+  (_: AppState, props: { resourceIds?: string[] }) => props.resourceIds || "",
+  (ops, search, envId, resourceIds): DeployActivityRow[] => {
+    if (search === "" && envId === "" && resourceIds.length === 0) {
```
madhuravius, line 70: not sure if you want to do this, but i think we should so we can get resources and its possible children (app -> services)

src/deploy/service/index.ts
```diff
@@ -167,20 +168,66 @@ const initService = defaultDeployService();
...
   "/accounts/:id/services",
 );
 export const fetchAppServices = api.get<{ id: string }>("/apps/:id/services");
 
+export const fetchServiceOperations = api.get<{ id: string }>(
+  "/services/:id/operations",
+  api.cache(),
+);
+export const cancelServicesOpsPoll = createAction("cancel-services-ops-poll");
+export const pollServiceOperations = api.get<{ id: string }>(
```
madhuravius, line 190: i want to combine this with app for the activity page for apps somehow (1 app + N services api calls) per app, or we can alternatively just query on account (?)

src/ui/shared/active-operation-notice.tsx
```diff
@@ -37,7 +37,7 @@ export const ActiveOperationNotice = ({
   resourceType,
 }: {
   resourceId: string;
-  resourceType: Extract<ResourceType, "app" | "database">;
+  resourceType: Extract<ResourceType, "app" | "database" | "service">;
```
madhuravius, line 40: revert

src/deploy/service/index.ts
madhuravius: not doing this now, for later

src/deploy/activity/index.ts
madhuravius: this is okay, we're gonna do this

src/ui/pages/app-detail-service-scale.tsx
madhuravius: this is done by loader

src/deploy/app/utils.ts
```diff
@@ -82,6 +101,14 @@ export const CONTAINER_PROFILES = {
     maximumContainerCount: 32,
   },
 };
+export const CONTAINER_PROFILE_TYPES: string[] =
+  Object.keys(CONTAINER_PROFILES);
+export const EXPONENTIAL_CONTAINER_SIZES_BY_PROFILE = (
+  profile: ContainerProfileTypes,
+): number[] | undefined =>
```
neurosnap, line 108: I'm curious, when would this be undefined?  The filter should always return an array.  What might be `undefined` is an element inside the array.  So this would be better expressed as `(number | undefined)[]`.  
  But that also isn't ideal so I would append a `.filter(Boolean)` so we are guaranteed to get `number[]`.

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -0,0 +1,333 @@
...
+            {service ? (
+              <div className="my-4 flex justify-between">
+                <div>
+                  <Label>Pricing</Label>
+                  <p className="text-gray-500">
+                    {service.containerCount} container
+                    {service.containerCount > 1 ? "s" : ""} x{" "}
+                    {service.containerMemoryLimitMb / 1024} GB x $
+                    {(
+                      CONTAINER_PROFILES[service.instanceClass]
```
neurosnap, line 222: I'd move this into a variable up-top

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -0,0 +1,333 @@
...
+                    per GB/hour
+                  </p>
+                </div>
+                <div>
+                  <p className="text-gray-500">
+                    Current Estimated Monthly Cost
+                  </p>
+                  <p className="text-right text-lg text-green-400">
+                    $
+                    {(
```
neurosnap, line 234: I'd move this into a variable up-top

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -0,0 +1,333 @@
...
+            {changesExist ? (
+              <div className="my-4 flex justify-between">
+                <div>
+                  <Label>Pricing</Label>
+                  <p className="text-gray-500">
+                    {containerCount || 1} container
+                    {containerCount > 1 ? "s" : ""} x {containerSize / 1024} GB
+                    x $
+                    {(
+                      CONTAINER_PROFILES[containerProfileType]
```
neurosnap, line 284: I'd move this into a variable up-top

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -0,0 +1,333 @@
...
+                    ).toFixed(2)}{" "}
+                    per GB/hour
+                  </p>
+                </div>
+                <div>
+                  <p className="text-gray-500">New Estimated Monthly Cost</p>
+                  <p className="text-right text-lg text-green-400">
+                    $
+                    {(
+                      computedCostsForContainer(
```
neurosnap, line 295: I'd move this into a variable up-top

src/ui/pages/db-detail-scale.tsx
```diff
@@ -115,28 +202,133 @@ export const DatabaseScalePage = () => {
...
+                    {database.disk.size} GB disk x $0.20 per GB/month
+                  </p>
+                ) : null}
+              </div>
+              <div>
+                <p className="text-gray-500">Estimated Monthly Cost</p>
+                <p className="text-right text-lg text-green-400">
+                  $
+                  {(
+                    computedCostsForContainer(
```
neurosnap, line 238: I'd move this into a variable up-top

src/ui/pages/db-detail-scale.tsx
```diff
@@ -46,37 +110,53 @@ export const DatabaseScalePage = () => {
...
+                      }}
                     >
-                      <option value="test" disabled>
-                        General Purpose (M)
-                      </option>
+                      {CONTAINER_PROFILE_TYPES.map((containerProfileType) => (
+                        <option value={containerProfileType}>
+                          {
+                            CONTAINER_PROFILES[
+                              containerProfileType as ContainerProfileTypes
```
neurosnap, line 128: Because we are casting here we lose all type safety.  Honestly I wouldn't reference these array variables directly and instead use a lookup function that guarantees we return valid data (even if what it returns is 0 or empty strings).

neurosnap (changes requested):
Overall looks good.  The main change I'd like to see is guaranteeing these lookup tables return values so there isn't as much ambiguity whether a value will be available or not.  I also think the two forms have a lot of calculations being computed inside the html and a lot of `.toFixed()`.  Moving all of that into properly labelled variables would help with readability.

Further, using functions for these lookups instead of accessing arrays directly would help provide safety when a lookup yields no value.

src/deploy/app/utils.ts
madhuravius: this was to make the type checker pass, i think it thinks that you can pass in a key that isn't possible somehow, i don't think it's smart enough to know that

madhuravius: this is now no longer needed

src/ui/pages/db-detail-scale.tsx
madhuravius: how would you write this without an `as`? i'm not sure how you can bypass a string => literal

madhuravius: this comment doesn't fully matter here, right? it's done in a select statement (we control this)

either way i did best effort by forcing an array of strings and adding needless casting in the on change

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -0,0 +1,343 @@
...
+                      className="mb-2 w-full appearance-none block px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-gray-500 focus:border-gray-500 sm:text-sm"
+                      placeholder="select"
+                      onChange={(e) => {
+                        e.preventDefault();
+                        setContainerSize(parseInt(e.target.value));
+                      }}
+                    >
+                      {EXPONENTIAL_CONTAINER_SIZES_BY_PROFILE(
+                        containerProfileType,
+                      )?.map((containerSizeOption) => (
```
neurosnap, line 216: Shouldn't need optional chaining here

src/ui/pages/db-detail-scale.tsx
```diff
@@ -115,28 +232,104 @@ export const DatabaseScalePage = () => {
...
+            <div className="flex mt-4">
+              <Button
+                className="w-40 mb-4 flex font-semibold"
+                disabled={!changesExist}
+                onClick={(e) => onSubmitForm(e)}
+                type="submit"
+              >
+                Save Changes
+              </Button>
+              {service && changesExist ? (
```
neurosnap, line 320: service will always exist

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -0,0 +1,343 @@
...
+                    <select
+                      value={containerSize}
+                      className="mb-2 w-full appearance-none block px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-gray-500 focus:border-gray-500 sm:text-sm"
+                      placeholder="select"
+                      onChange={(e) => {
+                        e.preventDefault();
+                        setContainerSize(parseInt(e.target.value));
+                      }}
+                    >
+                      {EXPONENTIAL_CONTAINER_SIZES_BY_PROFILE(
```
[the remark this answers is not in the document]
madhuravius: shouldn't we add a lint rule for this? seems like a social convention, happy to oblige but just an autoformattable thing right? or lint guard

src/ui/pages/db-detail-scale.tsx
```diff
@@ -1,36 +1,124 @@
...
+  ).toFixed(2);
+  const estimatedPricePerHour = (
+    requestedContainerProfile.costPerContainerHourInCents / 100
+  ).toFixed(2);
+  const estimatedPrice = (
+    computedCostsForContainer(1, requestedContainerProfile, containerSize)
+      .estimatedCostInDollars /
+      1000 +
+    diskValue * 0.2
+  ).toFixed(2);
```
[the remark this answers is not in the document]
madhuravius: can do in followup if needed with tests, but would just be a refactor of these pages

neurosnap (approved):

PR #254, opened 2023-08-09
feat: deprovision environment

neurosnap:
feat: change name for environment
Screenshot 2023-08-09 at 3 29 32 PM

### DIFF
--- a/src/ui/pages/environment-detail-settings.tsx
+++ b/src/ui/pages/environment-detail-settings.tsx
@@ -30,13 +30,10 @@ import {
   Input,
 } from "../shared";
 import { environmentsUrl } from "@app/routes";
+import { handleValidator } from "@app/validator";
 
 const validators = {
-  name: (name: string) => {
-    if (name.trim() === "") {
-      return "Environment cannot be empty";
-    }
-  },
+  handle: handleValidator,
 };
 
 const EnvChangeName = ({ envId }: { envId: string }) => {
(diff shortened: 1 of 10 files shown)


src/ui/pages/environment-detail-settings.tsx
```diff
@@ -1,122 +1,245 @@
...
+  const env = useSelector((s: AppState) =>
+    selectEnvironmentById(s, { id: envId }),
   );
 
-  useEffect(() => {
-    setHandle(environment.handle);
-  }, [environment.id]);
+  const [confirm, setConfirm] = useState(false);
+  const [handle, setHandle] = useState<string>("");
+  const [errors, validate] = useValidator<string, typeof validators>(
```
madhuravius, line 50: are you using the handle regex match as well (?) i swear we had one

madhuravius: https://github.com/aptible/app-ui/blob/main/src/ui/shared/db/creator-form.tsx#L59 this one

neurosnap: i didn't see any validators on the environment -- will double check

madhuravius: figured it was the common handle validator we use everywhere else (some kind of reusable func/decorator thing)

joeclayallday:
Tiny thing - we don't show a label of Confirm Name anywhere for these deprovision blocks, so maybe we remove that to pattern match.
Screen Shot 2023-08-09 at 7 18 09 PM

src/ui/pages/environment-detail-settings.tsx
```diff
@@ -1,122 +1,245 @@
...
-              {isDeleting ? "Deleting..." : "Delete Environment"}
-            </Button>
-          </div>
+      </div>
+
+      <form onSubmit={onSubmit}>
+        <div>
+          This will permanently deprovision <strong>{env.handle}</strong>{" "}
+          environment. This action cannot be undone. If you want to proceed,
+          type the <strong>{env.handle}</strong> below to continue.
```
joeclayallday, line 195: We should remove the word "the" here on line 195

neurosnap: done, thanks

madhuravius (approved):

PR #259, opened 2023-08-10
feat: cert detail

neurosnap:
feat: cert delete
Screenshot 2023-08-10 at 3 40 33 PM
 

https://github.com/aptible/app-ui/assets/1940365/bdff59a0-d43f-4f64-9c0c-a26027dc6562

https://github.com/aptible/app-ui/assets/1940365/3544d30a-f538-4d24-8a58-64994156bd47

### DIFF
--- a/src/ui/pages/cert-detail-endpoints.tsx
+++ b/src/ui/pages/cert-detail-endpoints.tsx
@@ -19,5 +19,5 @@ export const CertDetailEndpointsPage = () => {
       envId: cert.environmentId,
     }),
   );
-  return <EndpointList exposedPorts={[]} endpoints={endpoints} />;
+  return <EndpointList endpoints={endpoints} />;
 };
--- a/src/ui/pages/cert-detail-endpoints.tsx
+++ b/src/ui/pages/cert-detail-endpoints.tsx
@@ -1,0 +1,23 @@
+import { EndpointList } from "../shared";
+import {
+  fetchEndpointsByCertId,
+  selectCertificateById,
+  selectEndpointsByCertificateId,
+} from "@app/deploy";
+import { AppState } from "@app/types";
+import { useSelector } from "react-redux";
+import { useParams } from "react-router";
+import { useQuery } from "saga-query/react";
+
+export const CertDetailEndpointsPage = () => {
+  const { id = "" } = useParams();
+  useQuery(fetchEndpointsByCertId({ certId: id }));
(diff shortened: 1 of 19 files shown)


src/ui/pages/cert-detail-settings.tsx
```diff
@@ -0,0 +1,107 @@
...
+      <h1 className="text-lg text-red-500 font-semibold">
+        <IconAlertTriangle
+          className="inline pr-3 mb-1"
+          style={{ width: 32 }}
+          color="#AD1A1A"
+        />
+        Delete Certificate
+      </h1>
+      <form className="mt-4 flex flex-col gap-4" onSubmit={onSubmit}>
+        {endpoints.length > 0 ? (
```
madhuravius, line 66: i'm brain farting but i cannot see the loader banner that we use sometimes here. is that now implicit somewhere (or referenced in another sibling container)

the certificate delete could return an error from the backend and we should present that to the user if that's the case, right? (i can see it erroring if the frontend has incorrect data for any reason or something else is preventing deletion for some reason)

usually it looks like this i think:

```tsx
{loader.isError ? <BannerMessages {...loader} /> : null}
```

src/ui/pages/cert-detail-endpoints.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { id = "" } = useParams();
+  useQuery(fetchEndpointsByCertId({ certId: id }));
+  const cert = useSelector((s: AppState) => selectCertificateById(s, { id }));
+  const endpoints = useSelector((s: AppState) =>
+    selectEndpointsByCertificateId(s, {
+      certificateId: id,
+      envId: cert.environmentId,
+    }),
+  );
+  return <EndpointList exposedPorts={[]} endpoints={endpoints} />;
```
madhuravius, line 22: if exposedports is optional why not just default it to a `[]` on the component? (personal pref?)

neurosnap: Removed `exposedEndpoints` entirely because of this discussion: https://aptible.slack.com/archives/C022BUSK487/p1691694149586049

src/ui/pages/cert-detail-settings.tsx
neurosnap: Good catch, added.

madhuravius (approved):

PR #265, opened 2023-08-14
Sc 14103 metric drains

neurosnap:


src/ui/pages/create-log-drain.tsx
```diff
@@ -0,0 +1,30 @@
+import { EnvironmentDetailLayout } from "../layouts";
+import { EnvironmentSelect, SelectOption } from "../shared";
+import { useState } from "react";
+import { useSearchParams } from "react-router-dom";
+
+export const CreateLogDrainPage = () => {
+  const [params] = useSearchParams();
+  const queryEnvId = params.get("environment_id") || "";
+  const [envId, setEnvId] = useState(queryEnvId);
+  console.log(envId);
```
madhuravius, line 10: prob wanna remove

src/ui/pages/create-metric-drain.tsx
```diff
@@ -0,0 +1,345 @@
...
+              ariaLabel="Type"
+              id="metric-type"
+              options={options}
+              onSelect={onTypeSelect}
+              value={drainType}
+            />
+          </FormGroup>
+
+          {drainType === "influxdb_database" ? (
+            <FormGroup
```
madhuravius, line 193: nit but you probably could just move this to a different set of components

src/ui/pages/environment-detail-integrations.tsx
```diff
@@ -268,11 +271,26 @@ const MetricDrainsSection = ({ id }: { id: string }) => {
 
 export const EnvironmentIntegrationsPage = () => {
   const { id = "" } = useParams();
+  const navigate = useNavigate();
+  const onCreateMetrics = () => {
+    navigate(createMetricDrainUrl(id));
+  };
+  const onCreateLogs = () => {
```
madhuravius, line 278: same as above or i can drop it when i pick this guy up

src/ui/pages/create-metric-drain.tsx
```diff
@@ -0,0 +1,345 @@
...
+                  type="text"
+                  id="database"
+                  value={database}
+                  onChange={(e) => setDb(e.currentTarget.value)}
+                />
+              </FormGroup>
+            </>
+          ) : null}
+
+          <ButtonCreate
```
madhuravius, line 334: you probably want to add the loader banner messages here too, right? if the backend errors?

src/ui/pages/create-metric-drain.tsx
```diff
@@ -0,0 +1,345 @@
...
+                    </ExternalLink>{" "}
+                    (or reuse an existing one) and paste it here.
+                  </p>
+                </div>
+              }
+              feedbackMessage={errors.apiKey}
+              feedbackVariant={errors.apiKey ? "danger" : "info"}
+            >
+              <Input
+                type="text"
```
madhuravius, line 238: should probably be a password/sensitive type, no?

src/ui/pages/create-metric-drain.tsx
```diff
@@ -0,0 +1,345 @@
...
+  },
+  // datadog
+  apiKey: (p: CreateMetricDrainProps) => {
+    if (p.drainType !== "datadog") return;
+    if (p.apiKey === "") return "Must provide a Datadog API key";
+  },
+};
+
+export const CreateMetricDrainPage = () => {
+  const navigate = useNavigate();
```
madhuravius, line 76: not sure if this happens already but you might wanna do the fetch for databases (in case someone loads this page from a hyper link) if it doesn't load already (for the influxdb env dropdown)

src/ui/pages/create-metric-drain.tsx
```diff
@@ -0,0 +1,345 @@
...
+  const dispatch = useDispatch();
+  const [params] = useSearchParams();
+  const queryEnvId = params.get("environment_id") || "";
+  const [envId, setEnvId] = useState(queryEnvId);
+  const [dbId, setDbId] = useState("");
+  const [handle, setHandle] = useState("");
+  const [apiKey, setApiKey] = useState("");
+  const [protocol, setProtocol] = useState<"http" | "https">("https");
+  const [hostname, setHostname] = useState("");
+  const [port, setPort] = useState("");
```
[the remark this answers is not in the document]
neurosnap: For other ports we use string for input value.  Maybe that's wrong?  I'll leave it as-is for consistency but we might want to revisit.

src/ui/pages/create-metric-drain.tsx
neurosnap: Hmm, it's a valid point but we don't currently do it in `deploy-ui` so I'm not sure it's really an issue.

src/ui/pages/create-metric-drain.tsx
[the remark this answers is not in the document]
madhuravius: we can, it tripped me up for some number fields i was working with on scaling, which is hwy i noticed/raised

madhuravius (approved):

PR #271, opened 2023-08-16
chore: added unit tests and added container profile selection

madhuravius:
timeboxed, but was unable to get services to get picked up in the msw interceptor (logs is also erroring randomly)


src/app/test/app-detail-service-scale.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  waitForToken,
+} from "@app/test";
+import { rest } from "msw";
+
+describe("AppDetailServiceScalePage", () => {
+  it("should successfully show app service scale page happy path", async () => {
+    server.use(
+      ...verifiedUserHandlers(),
+      ...stacksWithResources({ accounts: [testAccount], apps: [testApp] }),
+      rest.get(`${testEnv.apiUrl}/operations/:id/logs`, (_, res, ctx) => {
```
madhuravius, line 29: this is erroring but not failing somehow

src/app/test/db-detail-scale.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { rest } from "msw";
+
+describe("DatabaseScalePage", () => {
+  it("should successfully show database scale page happy path", async () => {
+    server.use(
+      ...verifiedUserHandlers(),
+      ...stacksWithResources({
+        accounts: [testAccount],
+        databases: [testDatabasePostgres],
+        services: [testServicePostgres],
```
madhuravius, line 31: this service does not get picked up

src/app/test/app-detail-service-scale.test.tsx
neurosnap: What is erroring about it?  Is `saga-query` erroring?  If so it might be because you aren't returning `JSON` which we expect.

src/app/test/db-detail-scale.test.tsx
neurosnap: We don't fetch services directly -- as far as I know.  You want to load `apps: [testApp]` here and then the service will get picked up.

src/app/test/app-detail-service-scale.test.tsx
madhuravius: resolved with huddle, basically need to stub on the `stackWithResources()` or other predecessor in tree for server.use

neurosnap (approved):

PR #275, opened 2023-08-16
[WIP]Sign Up Page Tweaks

joeclayallday:
Responsive Sign Up page with feedback from marketing

Screen Shot 2023-08-18 at 10 18 25 PM

Screen Shot 2023-08-18 at 10 10 35 PM


src/app/router.tsx
```diff
@@ -587,6 +588,11 @@ export const appRoutes: RouteObject[] = [
     ],
   },
 
+  {
```
madhuravius, line 591: if you comment this or delete these lines we can just merge this progress

joeclayallday: Still waiting on feedback from Tin. I added back in the removed sign up routing stuff. I can revert that commit OR we can keep this open before launch.

PR #281, opened 2023-08-18
feat: create db endpoint

neurosnap:
Screenshot 2023-08-18 at 3 03 45 PM


madhuravius (approved):
LGTM, they should always be able to edit when viewing on settings pane

PR #286, opened 2023-08-22
Top-level Endpoints

neurosnap:
Screenshot 2023-08-21 at 11 13 38 PM


src/ui/shared/endpoint/endpoint-list.tsx
```diff
@@ -0,0 +1,336 @@
...
+  );
+};
+
+const headers = [
+  "Hostname",
+  "Resource",
+  "Status",
+  "Placement",
+  "Platform",
+  "ACME Status",
```
madhuravius, line 149: are certs embedded? if not we probably need to add one MORE call at the top level or this might not load initially

neurosnap: I don't think they are, good idea.

src/deploy/endpoint/index.ts
```diff
@@ -227,32 +238,211 @@ export const selectFirstEndpointByAppId = createSelector(
...
+    return enps.filter((enp) => {
+      const foundDbId = serviceToDbMap[enp.serviceId];
+      if (foundDbId !== dbId) return false;
+      const searchMatch = computeSearchMatch(enp, search);
+      return searchMatch;
+    });
+  },
+);
+
+export const selectEndpointsByCertIdForTableSearch = createSelector(
```
madhuravius, line 435: not important now - it's gonna suck but we need to make a followup ticket for all these selectors

neurosnap: Yeah, I'd like to write some tests because the lookups for these are a bit gnarly.

madhuravius (approved):
LGTM - only one possible issue (certificates may not be hydrated) and one followup (test coverage)

i don't think either of the above are urgent

madhuravius (approved):

madhuravius (changes requested):
broken environment page, from slack history:

image

image

`instrument.ts:132 The above error occurred in the <EnvironmentPageHeader> component:`

sry to drag my feet on this one a bit

madhuravius (approved):

PR #296, opened 2023-08-23
fix: create signup billing records 

neurosnap:


madhuravius (approved):
lgtm - i think we have to test this against staging to see how it actually is meant to play, as some of the 2fa parts are really hard to read as a person (we have to see how it will interact in practice and localhost doesn't seem to work for me)

left minor comments in slack

madhuravius (approved):

PR #302, opened 2023-08-24
chore: added restart actions for dbs/apps/drains

madhuravius:
~TODO - disable on v4 log drains as well (restart not allowed)~

---

* ~proof~
* ~tests~ (adding to part 2 with deletes)

image

image

image

delete works w/ confirm

image

image

image

if v4, would look like this:

image

### DIFF
--- a/src/ui/pages/app-detail-settings.tsx
+++ b/src/ui/pages/app-detail-settings.tsx
@@ -3,7 +3,6 @@ import {
   Box,
   BoxGroup,
   Button,
-  ButtonIcon,
   ButtonLinkExternal,
   ExternalLink,
   FormGroup,
--- a/src/ui/pages/app-detail-settings.tsx
+++ b/src/ui/pages/app-detail-settings.tsx
@@ -3,13 +3,16 @@ import {
   Box,
   BoxGroup,
   Button,
+  ButtonIcon,
   ButtonLinkExternal,
   ExternalLink,
   FormGroup,
   IconAlertTriangle,
   IconExternalLink,
+  IconRefresh,
   IconTrash,
   Input,
+  Label,
   PreCode,
   listToInvertedTextColor,
 } from "../shared";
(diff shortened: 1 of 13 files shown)


joeclayallday:
Pushed up a small design tweak @madhuravius 
Screen Shot 2023-08-23 at 8 39 06 PM

src/ui/pages/app-detail-settings.tsx
```diff
@@ -76,6 +79,16 @@ export const AppSettingsPage = () => {
     navigate(appActivityUrl(id));
   };
 
+  const restartAction = restartApp({ id });
+  const restartLoader = useLoader(restartAction);
+  const submitRestart = (e: SyntheticEvent) => {
```
neurosnap, line 84: Use the actual event type here.

src/ui/pages/app-detail-settings.tsx
```diff
@@ -157,15 +170,32 @@ export const AppSettingsPage = () => {
...
+              disabled={
+                isUpdating ||
+                updatingAppLoader.isLoading ||
+                restartLoader.isLoading
+              }
             >
-              {isUpdating || updatingAppLoader.isLoading
+              {isUpdating ||
+              updatingAppLoader.isLoading ||
+              restartLoader.isLoading
```
neurosnap, line 198: `Button` has `isLoading` which will: disable the button and set the "Loading ..." text.

src/ui/pages/app-detail-settings.tsx
```diff
@@ -157,15 +170,32 @@ export const AppSettingsPage = () => {
...
+            {isDeprovisioning ? "Restarting..." : "Restart"}
+          </Button>
+
           <hr className="mt-6" />
 
           <div className="flex mt-4">
             <Button
               className="w-40 flex semibold"
               type="submit"
-              disabled={isUpdating || updatingAppLoader.isLoading}
```
neurosnap, line 166: What is the purpose of `isUpdating`?  The loaders are what indicate if something is being updated so this seems unnecessary.

src/ui/pages/app-detail-settings.tsx
madhuravius: got it

src/ui/pages/app-detail-settings.tsx
madhuravius: this is an older component despite you and me working on update before loaders were as popular as they are now in this codebase (when you were OOO i think and i made stuff up as i went)

joeclayallday (approved):

almathew (approved):

PR #303, opened 2023-08-24
Add Estimated Cost to DB tables

joeclayallday:
Add Estimated Cost to DB tables

Screen Shot 2023-08-24 at 1 20 40 PM
Screen Shot 2023-08-24 at 1 21 25 PM


madhuravius (approved):
LGTM - only place there could be issues are mid deploy but i think we can test this

PR #307, opened 2023-08-25
DO NOT MERGE - feat: maintenance in ui

madhuravius:
Do not merge this, there's a lot of problems here I realized after building this. This is pointed to restarts/deletes base branch because it's just easier to point to the settings page for that.

Things that have to be resolved:

* I opted to mirror apps/dbs (in hindsight this was a very bad decision) with an added expensive field that requires us to store this in a separate place. I hacked the heck out of some middleware to make this work. do as you please but i would never merge this thing
* When writing this PR i figured I'd generalize app/dbs, but theoretically I think conflicts can occur in store and we could have mismatches and conflicts
* The tooltips/links all retrieve, but be very careful when hitting this endpoint. It can be expensive and it's part of the reason it wasn't attached to the main model, see Notion doc for design decisions/tradeoffs on Slack

---

Proof (this does work):

image

image

On hover:

image

Ex for dbs working/ proof: this list should correspond to the outputs displayed on THIS list (from CLI/backend  PR): 

image

### DIFF
--- a/src/deploy/app/index.ts
+++ b/src/deploy/app/index.ts
@@ -419,6 +419,30 @@ export const appEntities = {
   }),
 };
 
+export const restartApp = api.post<{ id: string }, DeployOperationResponse>(
+  ["/apps/:id/operations", "restart"],
+  function* (ctx, next) {
+    const { id } = ctx.payload;
+    const body = {
+      type: "restart",
+      id,
+    };
+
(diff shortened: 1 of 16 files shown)


neurosnap:
It is going to be faster for us to start from scratch on this PR instead of trying to resolve all of the conflicts, too much has changed since this was put up.  Closing for now.

PR #309, opened 2023-08-25
fix: login redirect

neurosnap:


madhuravius (approved):

madhuravius:
should find a way to test/mock this at some point, kinda scary to wing i this way but we should merge/release asap

PR #310, opened 2023-08-25
fix: use correct billing api url

neurosnap:


madhuravius (approved):
might wanna update readme and other places you see this

joeclayallday (approved):

PR #311, opened 2023-08-25
Make m5 the default m instance class

aaw:
Also allows selecting m5 in scaling panels. We're upgrading everyone to m5 starting next week so it's fine for people to use scaling to move apps and databases over now that we have full support for m5 as default.


src/deploy/service/index.ts
```diff
@@ -19,7 +19,7 @@ import { CONTAINER_PROFILES, GB } from "../container/utils";
 import { selectDeploy } from "../slice";
 import { poll } from "saga-query";
 
-export const DEFAULT_INSTANCE_CLASS: InstanceClass = "m4";
+export const DEFAULT_INSTANCE_CLASS: InstanceClass = "m5";
```
neurosnap, line 22: I moved all referenced of "m4" to use this variable.  If you could confirm we don't have any other magic strings that would be great!

aaw: Thanks, I merged your changes and I think I caught the final magic string in this PR

aaw:
https://app.shortcut.com/aptible/story/17597/support-m5-as-default

aaw:
@neurosnap friendly ping for a review

neurosnap (approved):

PR #347, opened 2023-09-01
Fix-Typo

joeclayallday:
Fix Typo
• Should say restart instead of reload

BEFORE
Screen Shot 2023-09-01 at 12 45 36 PM

AFTER
Screen Shot 2023-09-01 at 12 46 28 PM


almathew (approved):

almathew:
Thanks for grabbing this! I probably should've just done it rather than making the note...

PR #359, opened 2023-09-07
Database backup exclusion

benjodo:
https://app.shortcut.com/aptible/story/17406/allow-databases-to-be-individually-excluded-from-backup-policies

@neurosnap I'm a veeeeeery green FE hack, would appreciate you looking at this PR with skepticism 🙏

### DIFF
--- a/src/ui/pages/db-detail-settings.tsx
+++ b/src/ui/pages/db-detail-settings.tsx
@@ -134,7 +134,7 @@ const DatabaseNameChange = ({ database }: DbProps) => {
     [...logDrains, ...metricDrains] || [];
 
   return (
-    <form onSubmit={onSubmit}>
+    <form onSubmit={onSubmit} className="flex flex-col gap-4">
       <FormGroup label="Database Name" htmlFor="input-name">
         <Input
           name="app-handle"
@@ -178,6 +178,7 @@ const DatabaseNameChange = ({ database }: DbProps) => {
 
       <FormGroup label="Database Backups" htmlFor="input-backup">
         <select
+          className="border-black-100 text-black hover:border-black active:border-black-100 active:text-black disabled:bg-black-50 disabled:border-black-100 disabled:text-black rounded-md shadow-sm"
           name="database-backup"
           value={enableBackups.toString()}
           onChange={(e) => setEnableBackups(e.currentTarget.value === "true")}
@@ -194,7 +195,7 @@ const DatabaseNameChange = ({ database }: DbProps) => {
         </select>
       </FormGroup>
 
-      <Group variant="horizontal" size="sm" className="mt-4">
+      <Group variant="horizontal" size="sm">
         <ButtonCreate
           envId={database.environmentId}
           className="w-40 semibold"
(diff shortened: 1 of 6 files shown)


src/ui/pages/db-detail-settings.tsx
```diff
@@ -170,13 +176,34 @@ const DatabaseNameChange = ({ database }: DbProps) => {
         ) : null}
       </FormGroup>
 
+      <FormGroup label="Database Backups" htmlFor="input-backup">
+        <select
```
neurosnap, line 180: We have a `<Select />` component that keeps styles consistent, I would recommend importing that.

neurosnap (commented):
Looks pretty close to me!  Just to make sure: the database API endpoint does return this field `enabled_backups`?

benjodo:
> Looks pretty close to me! Just to make sure: the database API endpoint does return this field `enabled_backups`?

@neurosnap yes, the deploy-api change for `enable_backups` went live a few days back when you shipped it

joeclayallday:
@benjodo I shortened the text for ya

Screen Shot 2023-09-07 at 7 12 31 PM
Screen Shot 2023-09-07 at 7 12 34 PM

joeclayallday:
Yea we should probably use the Select component OR if its easier you can use radio buttons and import the Radio component.

Radio component example: https://github.com/aptible/app-ui/blob/891545117816c3de5479f449ef8a6a8abe4091d8/src/ui/pages/app-create-endpoint.tsx#L292

Select component example: https://github.com/aptible/app-ui/blob/891545117816c3de5479f449ef8a6a8abe4091d8/src/ui/pages/create-metric-drain.tsx#L227

For both components you'll need add imports and create consts

@benjodo

src/ui/pages/db-detail-settings.tsx
```diff
@@ -170,13 +176,33 @@ const DatabaseNameChange = ({ database }: DbProps) => {
         ) : null}
       </FormGroup>
 
-      <Group variant="horizontal" size="sm" className="mt-4">
+      <FormGroup label="Database Backups" htmlFor="input-backup">
+        <select
+          className="border-black-100 text-black hover:border-black active:border-black-100 active:text-black disabled:bg-black-50 disabled:border-black-100 disabled:text-black rounded-md shadow-sm"
```
joeclayallday, line 181: If you import the select component, you can remove this className and it will inherit the same styles.

joeclayallday (approved):
Looks great! Have Eric give the final say. 

Nitpicks
• If i change the dropdown value and click cancel, the dropdown doesnt change back to my original state.

benjodo:
> Nitpicks • If i change the dropdown value and click cancel, the dropdown doesnt change back to my original state (not a big deal).

Thanks for the bug catch, @joeclayallday I've updated the code to fix that ᕕ( ᐛ )ᕗ

src/ui/pages/db-detail-settings.tsx
```diff
@@ -104,11 +106,17 @@ interface DbProps {
...
   const [handle, setHandle] = useState<string>("");
   useEffect(() => {
     setHandle(database.handle);
+    setEnableBackups(database.enableBackups);
   }, [database.id]);
-  const action = updateDatabase({ id: database.id, handle });
+  const action = updateDatabase({
+    id: database.id,
+    handle,
+    enable_backups: enableBackups,
```
neurosnap, line 118: `enable_backups` should be camelCase here and then when creating the `body` we convert to snake_case.

neurosnap (approved):
🚀

PR #389, opened 2023-09-20
add-copy-git-remote-jc

joeclayallday:
Adds copy icon to git remotes on the app detail header page
• After click, the icon shows feedback
• added feedback to all existing copy icons

Screenshot 2023-09-19 at 8 38 31 PM

### DIFF
--- a/src/ui/layouts/app-detail-layout.tsx
+++ b/src/ui/layouts/app-detail-layout.tsx
@@ -22,7 +22,7 @@ import {
 } from "@app/routes";
 import { setResourceStats } from "@app/search";
 import type { AppState, DeployApp } from "@app/types";
-import { useEffect, useMemo } from "react";
+import { SyntheticEvent, useEffect, useMemo } from "react";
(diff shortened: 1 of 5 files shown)


src/ui/layouts/app-detail-layout.tsx
```diff
@@ -64,7 +71,19 @@ export function AppHeader({ app }: { app: DeployApp }) {
...
+            <div className="flex flex-row items-center">
+              {app.gitRepo}
+              <Tooltip text="Copy">
+                <IconCopy
+                  variant="sm"
+                  className="ml-2 active:opacity-50"
+                  color="#888C90"
+                  onClick={(e) => handleCopy(e, `${app.gitRepo}`)}
+                />
+              </Tooltip>
```
neurosnap, line 84: This should be in its own shared component.  Let me take a stab at this real quick and then you can use it everywhere.

neurosnap (changes requested):

PR #390, opened 2023-09-20
feat: <CopyText /> component

neurosnap:
https://github.com/aptible/app-ui/assets/1940365/f610b743-ffa9-4296-89a1-49cdf85bf24b

### DIFF
--- a/src/ui/shared/copy.tsx
+++ b/src/ui/shared/copy.tsx
@@ -1,32 +1,33 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { IconCopy } from "./icons";
 import { Tooltip } from "./tooltip";
 
+const copy = async (text: string) => {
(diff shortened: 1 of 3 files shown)


joeclayallday (approved):

src/ui/shared/copy.tsx
```diff
@@ -0,0 +1,44 @@
...
+    });
+  };
+
+  return (
+    <div className="flex items-center gap-2">
+      <div>{text}</div>
+      <Tooltip text={success ? "Copied!" : "Copy"}>
+        <IconCopy
+          variant="sm"
+          className="ml-2 active:opacity-50"
```
joeclayallday, line 37: i think we can remove the ml-2 because we have the gap-2 above

joeclayallday (approved):

PR #406, opened 2023-09-22
fix database endpoint

grrowe:


src/deploy/endpoint/index.ts
```diff
@@ -211,7 +211,7 @@ export const selectEndpointsByServiceIds = createSelector(
   selectEndpointsAsList,
   (_: AppState, p: { ids: string[] }) => p.ids,
   (endpoints, serviceIds) => {
-    return endpoints.filter((end) => serviceIds.includes(end.serviceId));
+    return endpoints?.filter((end) => serviceIds?.includes(end.serviceId));
```
neurosnap, line 214: endpoints should always exist as well as serviceIds so we can remove the `?.`

joeclayallday (approved):

joeclayallday:
Looks good!
Screenshot 2023-09-26 at 3 43 20 PM

src/ui/shared/endpoint/endpoint-list.tsx
```diff
@@ -365,3 +366,47 @@ export function EndpointsByCert({ certId }: { certId: string }) {
...
+  dbId,
+}: { serviceId: string; dbId: string }) {
+  const [params, setParams] = useSearchParams();
+  const search = params.get("search") || "";
+  const onChange = (nextSearch: string) => {
+    setParams({ search: nextSearch });
+  };
+  const db = useSelector((s: AppState) => selectDatabaseById(s, { id: dbId }));
+  const endpoints = useSelector((s: AppState) =>
+    selectEndpointsByServiceIds(s, { ids: [serviceId], db, search }),
```
neurosnap, line 381: looks like we are just passing a single service id here, could we refactor `selectEndpointsByServiceIds` to be singular?

src/deploy/endpoint/index.ts
```diff
@@ -261,9 +254,11 @@ export const selectEndpointsByOrgAsList = createSelector(
   selectEndpointsAsList,
   selectEnvironmentsByOrgAsList,
   (envToServiceMap, enps, envs) => {
-    return enps.filter((enp) =>
```
neurosnap, line 264: Hmm, I can't remember why I wrote this.  I might need to double check tomorrow.

neurosnap (changes requested):

src/ui/shared/endpoint/endpoint-list.tsx
grrowe: Yes, will do.

src/deploy/endpoint/index.ts
```diff
@@ -261,9 +254,11 @@ export const selectEndpointsByOrgAsList = createSelector(
...
-      envs.some((env) => envToServiceMap[env.id]?.has(enp.serviceId)),
-    );
+    return enps;
```
neurosnap, line 257: This will return all endpoints which isn't what we want here.  We want endpoints related to an organization, which requires some hoop jumping.  Happy to pair through this to figure out the best way to ensure app and db endpoints are returned.

grrowe: Reverted and fixed the `selectEnvToServicesMap`, it was filtering anything without an `appId` so databases got the boot.

neurosnap (approved):
Nice work!

PR #425, opened 2023-09-27
feat change redirect urls after deprovisioning apps and databases

joeclayallday:
After deprovisioning apps and databases go to the Environment / Activity tab


neurosnap (approved):
Thanks, I missed some!

PR #459, opened 2023-10-02
refactor selectActivityByApp to include services

grrowe:
Had a bug (https://aptible.zendesk.com/agent/tickets/49443) where the App's activity page would not show the services activity as well ie the scale operations from the service.

- Refactor `ActivityByApp` to include the App's services as well.

### DIFF
--- a/src/deploy/app/index.ts
+++ b/src/deploy/app/index.ts
@@ -1,5 +1,13 @@
 import { api, cacheMinTimer, thunks } from "@app/api";
-import { call, poll, select } from "@app/fx";
+import {
+  all,
+  call,
+  poll,
+  put,
+  select,
+  setLoaderStart,
+  setLoaderSuccess,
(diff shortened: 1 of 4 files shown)


joeclayallday (approved):

neurosnap (changes requested):
Unfortunately, this is more complicated because we aren't polling for service operations on this page so the data will be stale immediately.  We also can't easily poll each service individually so I think to do this the right way would be to ensure when fetching app operations we also get service operations.

neurosnap:
Continuing discussion in Slack: https://aptible.slack.com/archives/C022BUSK487/p1696269097698449

neurosnap:
Screenshot 2023-11-01 at 10 31 54 AM

neurosnap (approved):

joeclayallday (approved):

PR #479, opened 2023-10-04
add tuna metrics on support form

grrowe:


src/ui/pages/support.tsx
```diff
@@ -105,6 +106,13 @@ export const SupportPage = () => {
     e.preventDefault();
     const attachments = attachedFiles.map((file) => file.token);
     dispatch(createSupportTicket({ ...formState, attachments: attachments }));
+    
+    const w = window as any;
```
neurosnap, line 110:
```ts
import { tunaEvent } from "@app/tuna";
// ...
tunaEvent('submittedAppUiSupportForm', `{ "viewedDocs": "${viewedSuggestion}", "email": "${user.email}" }`);
```

src/ui/pages/support.tsx
```diff
@@ -236,7 +245,13 @@ export const SupportPage = () => {
                       return (
                         <li key={key + 1}>
                           <div>
-                            <a target="_blank" href={hit.url} rel="noreferrer">
+                            <a target="_blank" href={hit.url} rel="noreferrer" onClick={() => {
+                              const w = window as any;
+                              if (w.aptible?.event) {
+                                  setViewedSuggestion(true)
+                                  w.aptible.event('usedSupportSuggestion', `{ "suggestedUrl": "${hit.url}", "email": "${user.email}" }`);
```
neurosnap, line 252: same here

neurosnap (changes requested):
I just recently formalized how we use tuna in `app-ui` which ensures we use it properly.

src/ui/pages/support.tsx
neurosnap: I don't think the data needs to be a string either

grrowe: Needs to be a string.
image
First is as a string, second as an object

src/ui/pages/support.tsx
```diff
@@ -105,6 +107,10 @@ export const SupportPage = () => {
     e.preventDefault();
     const attachments = attachedFiles.map((file) => file.token);
     dispatch(createSupportTicket({ ...formState, attachments: attachments }));
+    tunaEvent(
+      "submittedAppUiSupportForm",
+      `{ "viewedDocs": "${viewedSuggestion}", "email": "${user.email}" }`,
```
neurosnap, line 112: `JSON.stringify({ viewedDocs, email })`

grrowe: This encodes with the strange `"{ \"suggestedUrl\": \"https://www.aptible.com/docs/influxdb-metric-drain\", \"email\": \"graham.rowe@aptible.com\" }"` as well as not being able to use JSON selectors (`event_value::json->>'email') in postgres.

neurosnap: got it, thanks for explanation!

joeclayallday (approved):

neurosnap (approved):

PR #486, opened 2023-10-05
refactor: forced redirects

neurosnap:
This PR attempts to address our flaky and error prone way we force users into specific flows.  In particular we need to accommodate the current redirects:

- User is not authenticated -> go to `/login`
- User is not verified -> go to `/verify`
- User has no payment method -> go to `/billing`

This turned into quite an exploration and I had multiple existential crises.  I tried multiple different ways to streamline our forced redirects and this current implementation was the one that won.

# How do forced redirects work?

We have a couple react components that will check for critical pieces of data.  If that data meets the requirements for a forced redirect, we render `<Navigate to={...} />`.

At the time of writing this we have 4 forced redirect components:

- `<AuthRequired />` -> checks for access token
- `<EmailVerifyRequired />` -> checks for `user.verified === true`
- `<PaymentMethodRequired />` -> checks for `billing_detail._links.payment_methods.href`
- `<ElevateRequired />` -> checks for an elevated token in our redux store

# Design Principle

We store the user's access token inside local storage.  This allows us to check for access token before we fetch `/current_token`.  While that request is in-flight and there is an access token in our store, we start rendering the app.  Only after `/current_token` returns `401` do we reset the store and force the user to `/login`.

This provides us with some unique features:
- We don't block the user if we already have an access token in our redux store
- We render the page they are requesting immediately, which means we try to fetch that data first before all of our other preloaded data (e.g. user goes to app-detail page, we fetch that specific app via API before we fetch preloaded data, making the page render faster).


joeclayallday (approved):

joeclayallday (approved):

joeclayallday:
Appreciate the detailed write up — nice work 👏

PR #507, opened 2023-10-13
fix(sso-cli): if token is empty, inform user and let them refetch token

neurosnap:
Screenshot 2023-10-13 at 6 50 59 PM


fancyremarker (approved):
Thanks!

PR #508, opened 2023-10-13
fix(ip-allowlist): permit cidr

neurosnap:


src/validator/index.ts
```diff
@@ -19,6 +25,7 @@ export function ipValidator(ips: string[]) {
 
   ips.forEach((ip) => {
     if (isIPv4(ip)) return;
+    if (isIPv4Cidr(ip)) return;
     errs.push(`${ip} is not a valid IPv4 address`);
```
fancyremarker, line 30: Should say "is not a valid IPv4 address or CIDR"

fancyremarker (approved):
Thanks again for the quick fix!

src/validator/index.ts
```diff
@@ -4,11 +4,18 @@ const IP_WHITELIST_MAX_SIZE = 25;
 const v4 =
   "(?:25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]\\d|\\d)(?:\\.(?:25[0-5]|2[0-4]\\d|1\\d\\d|[1-9]\\d|\\d)){3}";
 const v4exact = new RegExp(`^${v4}$`);
+// inlined from: https://github.com/silverwind/cidr-regex/blob/master/index.js#L4
+const v4Cidr = `${v4}\\/(3[0-2]|[12]?[0-9])`;
```
fancyremarker, line 8: Ish. For example, 8.8.8.1/24 is not actually valid (8.8.8.0/24 would be), but deploy-api handles this transformation so it's fine.

PR #539, opened 2023-10-25
refactor(table): simplify tables

neurosnap:


src/ui/pages/stacks.tsx
```diff
@@ -50,96 +55,84 @@ function StackListRow({ stack }: { stack: DeployStack }) {
...
-      />
-    </LoadResources>
+    <Group>
+      <Group size="sm">
+        <TitleBar description="Stacks represent the virtualized infrastructure where resources are deployed.">
+          Stacks
+        </TitleBar>
+
+        <FilterBar>
+          <div className="flex justify-between mb-1">
```
joeclayallday, line 95: Lets remove all these `mb-1`s from the FilterBar on this line only

src/ui/pages/stacks.tsx
```diff
@@ -50,96 +55,84 @@ function StackListRow({ stack }: { stack: DeployStack }) {
...
+
+            <ActionBar>
+              <ButtonLink to={createStackUrl()}>
+                <IconPlusCircle variant="sm" className="mr-2" /> New Dedicated
+                Stack
+              </ButtonLink>
+            </ActionBar>
+          </div>
+
+          <Group variant="horizontal" size="lg" className="items-center">
```
joeclayallday, line 110: And here add a `mt-1` on the Group that wraps the <DescBar> because every table has a DescBar but not every table has the filter bar

joeclayallday:
Can remove the space between Search and the ellipsis?
Screenshot 2023-10-25 at 3 50 35 PM

src/ui/pages/stacks.tsx
joeclayallday: Done

src/ui/pages/stacks.tsx
joeclayallday: Done

joeclayallday (approved):

joeclayallday (approved):

PR #555, opened 2023-11-01
[DoNotMerge]style(stacks): add outbound CIDR range to stack header

joeclayallday:
Todo:
• Make the CIDR Range values real (they currently don't show up)

Q) Do we also want to show VPC CIDR Range for the actual Aptible VPC Private IP range?

Screenshot 2023-11-01 at 1 55 34 PM
Screenshot 2023-11-01 at 1 55 38 PM


joeclayallday:
Maybe there's a way to do this with our copy command in a dropdown instead of making another detail-info-item.

PR #560, opened 2023-11-06
Invite members

neurosnap:
Features:
- View members list
- Invite members
- Pending invites
- Ability to change a user's roles
- Ability to remove a user from org


joeclayallday (approved):

joeclayallday:
Looks good, here's my suggestions:

• Should I be allowed to remove myself? 
• On the send invite page, change the button label to say "Share Invite"
• On the members list table, instead of a blue edit link, I'd use our table buttons instead. Pattern match off the databases Scale button on the databases list table
• Q) Do we wanna add in the download members list functionality in this PR?
• I think we can move the `Special search terms: verified and mfa. The inverse is also available: !verified and !mfa` into the top Members page tooltip and say: `Search members by name or status by typing verified, mfa, !verified, or !mfa`

src/ui/pages/settings-team-members.tsx
```diff
@@ -98,9 +97,11 @@ export const TeamMembersPage = () => {
                 {user.otpEnabled ? "Enabled" : "Disabled"}
               </Td>
               <Td variant="right">
-                <Button type="submit" size="sm">
-                  Edit
-                </Button>
+                {isOwner ? (
+                  <ButtonLink size="sm" to={teamMembersEditUrl(user.id)}>
```
joeclayallday, line 101: Add `className="w-fit justify-self-end inline-flex"` to get the button to look like this:

Screenshot 2023-11-14 at 9 56 05 AM

OR you can use `className="w-fit float-right"`

src/ui/pages/settings-team-pending-invites.tsx
```diff
@@ -1,3 +1,92 @@
...
+        <TBody>
+          {invitations.length === 0 ? (
+            <EmptyTr colSpan={2}>No pending invitations</EmptyTr>
+          ) : null}
+          {invitations.map((invite) => {
+            return (
+              <Tr key={invite.id}>
+                <Td>{invite.email}</Td>
+                <Td variant="right">
+                  <Group variant="horizontal" size="sm">
```
joeclayallday, line 65: Add `className="flex justify-end"` to float the buttons right
Screenshot 2023-11-14 at 10 04 56 AM

joeclayallday (approved):

PR #577, opened 2023-11-15
fix(db.create): look for existing db only in selected env

neurosnap:
When a user tries to create and provision a new database we first perform a check to see if a database with that exact name already exists within the system.  If it does already exist, we skip creating the database and instead move onto provision the database.

We do this because there are cases where a user can create a database but the provision operation fails (e.g. user hits plan limits).  When the user fixes the issue and returns to FTUX to try to reprovision, we would then say validation failed because the database already exists.  So if that database already exists we just try to reprovision.

Creating a database record and provisioning that database are two separate, atomic events.  Coordinating those events rests solely on the UI.  Users can create as many database records as they like, but provisioning a database then performs a plan limit check.

During FTUX the user selects a database and we inject a handle automatically based on env/db-type (the user can modify this).  When db provision fails (because of a plan limit), they have to bail out of the flow entirely to fix the issue.  When they come back to FTUX, their db is still sitting there with a failed provision operation.  At the time FTUX was built, this put the user in a broken state where they needed to delete the database (action we didnt support at the time) and essentially restart.

So instead we try to perform error recovery by looking up the database by handle and then triggering a provision operation on that db.

I will happily entertain the notion that this doesn't work well and we should rip it out for a better solution (e.g. let the user delete the database in the FTUX flow to recover), but this patch should resolve the issue.

https://aptible.slack.com/archives/C022RVCU2CT/p1700065685600039

### DIFF
--- a/src/app/test/create-db.test.tsx
+++ b/src/app/test/create-db.test.tsx
@@ -1,16 +1,22 @@
-import { act, fireEvent, render, screen } from "@testing-library/react";
-import userEvent from "@testing-library/user-event";
-import { rest } from "msw";
-
+import { defaultDatabaseResponse, defaultOperationResponse } from "@app/deploy";
+import { defaultHalHref } from "@app/hal";
 import {
+  createId,
   server,
   stacksWithResources,
   testAccount,
   testDatabaseOp,
+  testDatabaseServiceId,
+  testDestroyAccount,
+  testDisk,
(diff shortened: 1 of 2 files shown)


joshuaj194 (approved):
Tested locally and the fix works! Database is created in the expected environment and there is no impact to similarly named databases in other environments.

PR #582, opened 2023-11-15
Feat: add autoscaling column to services

joeclayallday:
**Todo**
• Wire up the logic to properly show if a Service is Enabled or Disabled

I'm guessing the logic will be something similar to this

`{user.otpEnabled ? "Enabled" : "Disabled"}`

But `user.otpEnabled` with your own code.

Mockup
1-App

### DIFF
--- a/src/ui/shared/app/services-detail.tsx
+++ b/src/ui/shared/app/services-detail.tsx
@@ -74,6 +74,12 @@ const ServiceListRow = ({
           </div>
         </Td>
 
+        <Td variant="center">
+          <div className={tokens.type.darker}>
+            Enabled
+          </div>
+        </Td>
+
         <Td variant="right">
           <Group size="sm" variant="horizontal">
             <ButtonLink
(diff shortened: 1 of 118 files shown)


joeclayallday:
I'm guessing the logic will be something similar to this

`{user.otpEnabled ? "Enabled" : "Disabled"}`

But `user.otpEnabled` with your own code.

aguilinger:
I think this will be sufficient?

almathew:
@joeclayallday happy to review if you want to move this out of draft

src/ui/shared/app/services-detail.tsx
```diff
@@ -37,6 +41,21 @@ const ServiceListRow = ({
...
+    selectEnvironmentById(s, { id: app.environmentId }),
+  );
+  const stack = useSelector((s: AppState) =>
+    selectStackById(s, { id: environment.stackId }),
+  );
+  const autoscaling = useCache<
+    HalEmbedded<{
+      service_sizing_policies: ServiceSizingPolicyResponse[];
+    }>
+  >(fetchServiceSizingPoliciesByServiceId({ service_id: service.id }));
```
neurosnap, line 54: This should probably be a hook so we can prevent a bunch of optional chaining everywhere.

neurosnap (changes requested):

neurosnap:
Closing in favor of: https://github.com/aptible/app-ui/pull/597

PR #587, opened 2023-11-17
v2

neurosnap:


src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -220,7 +229,7 @@ const VerticalAutoscalingSection = ({
                       id="maximum-memory"
                       name="maximum-memory"
                       type="number"
-                      value={nextPolicy.maximum_memory}
+                      value={nextPolicy.maximum_memory || ""}
```
neurosnap, line 232: move this check into `defaultServiceSizingPolicyResponse`

src/deploy/service/index.ts
```diff
@@ -289,7 +290,8 @@ export const defaultServiceSizingPolicyResponse = (
 ): ServiceSizingPolicyResponse => {
   const now = new Date().toISOString();
   return {
-    id: 1,
+    id: undefined,
+    service_id: "0",
```
neurosnap, line 294: empty string `""` is the canonical way to check for existence

aguilinger: ty!

src/ui/pages/app-detail-service-scale.tsx
aguilinger: I get errors if I set maximum_memory to empty string since it's supposed to be a number (but is nullable). If I leave it without the `|| ""` the console gets flooded with
Screenshot 2023-11-17 at 1 30 37 PM

PR #597, opened 2023-11-21
feat(app.services): vertical autoscaling column

neurosnap:
Supercedes: https://github.com/aptible/app-ui/pull/582

# BLOCKED

This PR is blocked by having a `deploy-api` endpoint to fetch all service sizing policies by App.  Without this, each row in the service list will have to fetch its own service sizing policy in order to show this autoscaling column (N+1 requests).


aguilinger (approved):

joeclayallday:
We've made lots of changes to the Services list tables. If we want to denote which services have Autoscale enabled I think this approach saves space and still communicates:

**Changes**
• Users can sort the actions column
• Show an A icon to denote Autoscale Enabled
• Show a tooltip that says "Autoscale Enabled" on hover

Screenshot 2023-12-19 at 1 53 09 PM

neurosnap:
Closing since it has been deprioritized.  See discussion: https://aptible.slack.com/archives/C0613V9E1PD/p1704383847457179

PR #627, opened 2023-12-08
style: add sorting arrow indicators

joeclayallday:
Add a icon to indicate column is sortable

https://github.com/aptible/app-ui/assets/4295811/91399447-dcc5-4727-999d-7839f1b1c54d


joshraker:
Would it be difficult to show an up+down arrow when not sorting and a down arrow when sorting to make it clear which column is being sorted on?

benjodo (approved):

PR #663, opened 2023-12-22
Fix: UTC formatting

joeclayallday:
Fix UTC Formatting on pages to show true UTC

**Shows in true UTC format so all dates match with our Logs and CLI**
Screenshot 2023-12-22 at 3 32 14 PM

**Shows the offset (ex. -06) so its more clear that this is local time** 
Screenshot 2023-12-22 at 3 31 59 PM

### DIFF
--- a/src/date/index.ts
+++ b/src/date/index.ts
@@ -90,7 +90,10 @@ export const prettyEnglishDate = (dateStr = "") => {
 };
 
 export const prettyUTCTime = (dateStr = "") => {
-  return new Date(dateStr).toISOString().slice(0, 19).replace('T', ' ') + (' UTC');
+  return `${new Date(dateStr)
+    .toISOString()
+    .slice(0, 19)
+    .replace("T", " ")} UTC`;
 };
 
 export const prettyEnglishDateWithTime = (dateStr = "") => {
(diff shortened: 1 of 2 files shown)


src/date/index.ts
```diff
@@ -90,7 +90,10 @@ export const prettyEnglishDate = (dateStr = "") => {
 };
 
 export const prettyUTCTime = (dateStr = "") => {
-  return format(isoToDate(dateStr), "yyyy-MM-dd HH:mm:ss 'UTC'");
+  return `${new Date(dateStr)
+    .toISOString()
+    .slice(0, 19)
```
fancyremarker, line 95: This is a little hacky, in that it assumes a very specific structure for the ISO output string. It should be safe given the detail of the ISO standard but can we find another way to do this, and/or add a test to ensure it's the value we want.

fancyremarker (changes requested):

src/date/index.ts
joeclayallday: I guess this always shows UTC — https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

PR #682, opened 2024-01-10
Prevent Disable VAS radio button from wiping settings

joshraker:
Now Enable/Disable will pause and resume scaling with the same settings instead of deleting the service's scaling policy membership.

Requires https://github.com/aptible/deploy-api/pull/1337 to be deployed first.

### DIFF
--- a/src/deploy/service/index.ts
+++ b/src/deploy/service/index.ts
@@ -439,7 +439,21 @@ export const updateServiceSizingPoliciesByServiceId = api.put<
 
 export const deleteServiceSizingPoliciesByServiceId = api.delete<{
   service_id: string;
-}>("/services/:service_id/service_sizing_policy");
+}>(["/services/:service_id/service_sizing_policy"], function* (ctx, next) {
+  yield* schema.update(db.loaders.start({ id: ctx.name }));
+  yield* next();
+
+  if (ctx.json.ok) {
+    yield* schema.update(
+      db.loaders.success({ id: ctx.name, message: "Policy changes saved" }),
+    );
+  } else {
+    const data = ctx.json.error as Error;
+    yield* schema.update(
+      db.loaders.error({ id: ctx.name, message: data.message }),
+    );
+  }
+});
 
 export const modifyServiceSizingPolicy =
   thunks.create<ServiceSizingPolicyEditProps>(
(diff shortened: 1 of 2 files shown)


joshraker:
https://app.shortcut.com/aptible/story/21590/allow-autoscaling-to-be-paused-from-the-ui

src/deploy/service/index.ts
```diff
@@ -448,26 +462,16 @@ export const modifyServiceSizingPolicy =
...
+          ),
+        );
       } else {
         updateCtx = yield* call(() =>
-          deleteServiceSizingPoliciesByServiceId.run(
-            deleteServiceSizingPoliciesByServiceId({
-              service_id: `${nextPolicy.service_id}`,
-            }),
+          updateServiceSizingPoliciesByServiceId.run(
+            updateServiceSizingPoliciesByServiceId(nextPolicy),
```
neurosnap, line 474: You can now remove  the second call to `updateServiceSizingPoliciesByServiceId`

joeclayallday:
Nice job! 👏

joshraker:
@joeclayallday I meant to add this screenshot for your review:

Screenshot 2024-01-10 at 10 49 36

src/deploy/service/index.ts
```diff
@@ -439,7 +439,21 @@ export const updateServiceSizingPoliciesByServiceId = api.put<
...
+    yield* schema.update(
+      db.loaders.success({ id: ctx.name, message: "Policy changes saved" }),
+    );
+  } else {
+    const data = ctx.json.error as Error;
+    yield* schema.update(
+      db.loaders.error({ id: ctx.name, message: data.message }),
+    );
+  }
+});
```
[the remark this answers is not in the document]
joshraker: Alright so I assume in your solution `yield* next()` performs or at least waits for the API request. I see how we pass along the success message but not any errors. Is that handled in `next()`?

src/deploy/service/index.ts
joshraker: I'm not sure which call you're referring to as "the second". Should it just be `updateServiceSizingPoliciesByServiceId(nextPolicy);`?

Does the same thing apply to `createServiceSizingPoliciesByServiceId`?

neurosnap: Exactly!

src/deploy/service/index.ts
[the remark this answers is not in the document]
neurosnap: Exactly!

`yield* next()` means "call the next middleware in the stack."  The middleware stack is exactly like `express` or `koa`.

The middleware stack lives here: https://github.com/aptible/app-ui/blob/main/src/api/index.ts#L256-L269

We have a loader middleware in the stack that automatically checked the status of a HTTP request and set the loader to be an error when the status is not ok: https://github.com/neurosnap/starfx/blob/f43c20f0880b1f4f971e04bd17d215a0700b8a7c/store/query.ts#L103

joeclayallday (approved):

joshraker:
With the API deployed the autoscaling can now be disabled while retaining settings.

PR #683, opened 2024-01-10
fix: create copy when using `array.sort()`

neurosnap:
`array.sort()` mutates the array in-place instead of returning a new array so we need to be careful because our store is frozen (immutable, read-only).


joeclayallday (approved):

joe-herman:
[related Slack thread](https://aptible.slack.com/archives/C05CMUW3LHX/p1704815260518129)
Thanks Eric!

joe-herman (approved):
Here's to hoping array copy+sorts don't tank performance

PR #684, opened 2024-01-12
chore(security): revoke other tokens for some ops

neurosnap:
- Changing password
- Updating MFA

https://aptible.slack.com/archives/C05CMUW3LHX/p1705072039725959?thread_ts=1704815260.518129&cid=C05CMUW3LHX

### DIFF
--- a/src/ui/pages/settings-security.tsx
+++ b/src/ui/pages/settings-security.tsx
@@ -2,7 +2,9 @@ import {
   fetchEmailVerificationPending,
   revokeAllTokens,
   revokeEmailVerification,
+  rmOtp,
 } from "@app/auth";
+import { updatePassword } from "@app/auth";
 import { prettyDateTime } from "@app/date";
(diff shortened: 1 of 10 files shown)


src/ui/pages/settings-security.tsx
```diff
@@ -95,6 +92,7 @@ const ChangePassword = () => {
     setError("");
   });
 
+  const isDisabled = pass === "";
```
joe-herman, line 95: It's been awhile for me: does this still dynamically update the button? I would think this needs to be a function

joe-herman (approved):
I think this looks mostly right, thank you for hammering this out so quickly!

joeclayallday (approved):

src/ui/pages/settings-security.tsx
neurosnap: React components constantly rerun so this will be computed whenever `pass` is updated.

PR #686, opened 2024-01-16
Show banner on scale page with last scale operation info

joshraker:
Before any scale operations have been created:

Screenshot 2024-01-16 at 17 24 58

While a scale operation is in progress:

Screenshot 2024-01-17 at 16 17 45

Scale operation succeeded:

Screenshot 2024-01-17 at 16 17 08

TODO:
- [x] Periodically fetch operations for the service to keep the banner up-to-date


joeclayallday (approved):

joeclayallday:
This is almost perfect, only note i have is to use our pretty UTC date formatter to remove the TZ letters from the timestamps:
`import { prettyDateTime } from "@app/date";`

joeclayallday (approved):

joshraker:
@joeclayallday I noticed that the height of the banner changes between progress and info because the `LoadingSpinner` defaults to the `sm` icon variant whereas icons normally default to the `base` variant. It was bothering me so I overrode the variant to `base` in `Banner` in order to keep the height the same. Let me know if you want me to change that back.

Screenshot 2024-01-17 at 10 19 09

Screenshot 2024-01-17 at 10 19 20

src/deploy/operation/index.ts
```diff
@@ -205,12 +205,26 @@ export const selectOperationsByEnvId = createSelector(
   (ops, envId) => ops.filter((op) => op.environmentId === envId),
 );
 
+export const findOperationsByServiceId = (
```
neurosnap, line 208: Non-blocker: if this isn't being used anywhere else it can be inlined inside the selector.

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -69,6 +75,54 @@ type AppScaleProps = {
   containerCount: number;
 };
 
+type LastScaleOperation = Pick<
+  DeployOperation,
+  "createdAt" | "containerCount" | "containerSize" | "status"
+>;
+
+const selectLastTwoScaleOps = createSelector(
```
neurosnap, line 83: selectors should live outside `ui` folder.

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -26,11 +30,13 @@ import {
   useSelector,
 } from "@app/react";
 import { appActivityUrl } from "@app/routes";
-import { HalEmbedded, InstanceClass } from "@app/types";
-import { SyntheticEvent, useEffect, useMemo, useState } from "react";
+import { DeployOperation, HalEmbedded, InstanceClass } from "@app/types";
+import { Fragment, SyntheticEvent, useEffect, useMemo, useState } from "react";
 import { useNavigate, useParams } from "react-router";
-import { useValidator } from "../hooks";
+import { createSelector } from "starfx/store";
```
neurosnap, line 36: Please use `@app/fx` for now

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -475,6 +529,43 @@ const VerticalAutoscalingSection = ({
   );
 };
 
+const LastScaleBanner = ({ serviceId }: { serviceId: string }) => {
+  const [lastScaleOp, olderScaleOp]: LastScaleOperation[] = useSelector((s) =>
```
neurosnap, line 533: You shouldn't need to cast to `LastScaleOperation`

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -475,6 +529,43 @@ const VerticalAutoscalingSection = ({
...
+  if (loader.isInitialLoading) {
+    return null;
+  }
+
+  return (
+    <Banner
+      variant={lastScaleComplete || !lastScaleOp ? "default" : "progress"}
+    >
+      {lastScaleOp ? (
+        <Fragment>
```
neurosnap, line 553: `<></>` is a shorthand version of `<Fragment>`

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -69,6 +75,54 @@ type AppScaleProps = {
...
+  selectNonFailedScaleOps,
+  (ops: LastScaleOperation[]) => {
+    let lastOp = ops[0];
+    if (lastOp == null) {
+      return ops;
+    }
+
+    // Remove the first operation so it's not considered when finding the past scale values
+    // Append an artificial operaiton with the default values for when a service is created
+    const opsWithDefault = ops.slice(1).concat({
```
neurosnap, line 93: `concat(defaultOperation(...))`

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -69,6 +75,54 @@ type AppScaleProps = {
...
+
+    // Copy the operations to avoid mutating the original
+    lastOp = { ...lastOp };
+    const oldOp = { ...opsWithDefault[0] };
+
+    // Scale operations don't have to have both a container count and size so
+    // we'll have to search past operations to find the last operation that
+    // modified each attribute
+    if (oldOp.containerCount == null) {
+      oldOp.containerCount =
```
neurosnap, line 108: I would break this apart for readability.

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -69,6 +75,54 @@ type AppScaleProps = {
...
+        opsWithDefault.find((op) => op.containerCount != null)
+          ?.containerCount || 0;
+    }
+
+    if (oldOp.containerSize == null) {
+      oldOp.containerSize =
```
neurosnap, line 114: I would break this apart for readability.

src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -475,6 +529,43 @@ const VerticalAutoscalingSection = ({
...
+  );
+  const lastScaleComplete = lastScaleOp?.status === "succeeded";
+  const action = pollServiceOperations({ id: serviceId });
+  const loader = useLoader(action);
+
+  const poller = useMemo(() => action, [serviceId]);
+  const cancel = useMemo(() => cancelServicesOpsPoll(), []);
+  usePoller({ action: poller, cancel });
+
+  if (loader.isInitialLoading) {
```
neurosnap, line 544: Could you explain the logic that you are trying to accomplish here?

joshraker:
https://app.shortcut.com/aptible/story/21535/adjust-test-order-grouping-to-improve-the-stability-of-tests

src/ui/pages/app-detail-service-scale.tsx
joshraker: 🤔 I'll probably have to rename it then so it's clear it's not actually the last two scale operations, but rather the current scale and previous scale of the service.

src/deploy/operation/index.ts
joshraker: I could see us using this in the nearish future to show service activity (i.e. scaling history) so I'm going to leave it.

joeclayallday:
@joshraker Banner icon changes look good! Thanks

src/ui/pages/app-detail-service-scale.tsx
joshraker: Good catch, I changed the selector definition and forgot to come back.

src/ui/pages/app-detail-service-scale.tsx
joshraker: `usePoller` - Poll the service's operations so that the banner will update in near-real time. That way, if someone is preparing to scale but someone else initiated one after the page loaded, they'll be notified of the scale in progress. Also, if a scale is in progress when the page loads, it will eventually update to the completed state.

Finally, we only load the first 250 operations from the organization so if we want to look at older scale operations, we need to fetch the service's operations. e.g. when I got up this morning, the service I'd scaled last night had never been scaled.

`loader.isInitialLoading` - Only render the banner once the initial load has finished. I tried `loader.isLoading` but that would remove and add the banner back each time the service's operations were polled. I'm also happy to render the banner with a loading message if that's preferred.

src/ui/pages/app-detail-service-scale.tsx
joshraker: Broken up into `selectServiceScale` and `selectPreviousServiceScale` and pulled some of the utility functionality out.

src/deploy/service/index.ts
```diff
@@ -325,6 +330,56 @@ export const selectAppByServiceId = createSelector(
...
+    });
+
+    return prev;
+  },
+);
+
+export const selectPreviousServiceScale = (
+  state: Parameters<typeof _selectPreviousServiceScale>[0],
+  { serviceId }: { serviceId: string },
+) => _selectPreviousServiceScale(state, { id: serviceId, serviceId });
```
joshraker, line 366: Is this the right way to "modify" the selector so that only `serviceId` needs to be passed in?

src/deploy/service/index.ts
```diff
@@ -325,6 +330,56 @@ export const selectAppByServiceId = createSelector(
...
+
+export const selectServiceScale = createSelector(
+  selectNonFailedScaleOps,
+  selectPreviousServiceScale,
+  (ops, prevOp) => {
+    const lastOps = ops.slice(0, 1).concat(prevOp);
+    const current: DeployOperation = { ...lastOps[0] };
+
+    scaleAttrs.forEach((attr) => {
+      (current as any)[attr] = findOperationValue(lastOps, attr);
```
joshraker, line 376: Without the cast, I was getting type errors here. Do you know of a way to make it work without the cast?

src/deploy/service/index.ts
joshraker: This is irrelevant because I realized that using the service values does not work for the previous scale but I'd still like the know the answer.

neurosnap: Every composed selector requires the props that are used for those selectors, there's no way around that.  It's hard to tell from this code what you are trying to do so if you'd like we could pair or chat through it.

src/deploy/service/index.ts
neurosnap: Does `forEach(attr: keyof DeployOperation`) work?

joshraker: Unfortunately, no.

```
yarn lint
src/deploy/service/index.ts:379:7 - error TS2322: Type 'any' is not assignable to type 'never'.

379       current[attr] = findOperationValue(lastOps, attr);
```

joeclayallday (approved):

src/deploy/service/index.ts
joshraker: I added a comment that I think explains it but if not happy to huddle and talk it through:

```
// selectServiceById expects the id prop whereas selectNonFailedScaleOps expects
// serviceId. Both of these should be the same value so accept serviceId and
// pass it in as both id and serviceId.
```

joeclayallday (approved):

src/deploy/service/index.ts
joshraker: Refactoring `selectNonFailedScaleOps` so that I don't have to do this based on our conversation.

neurosnap (approved):

PR #716, opened 2024-02-05
fix(settings): use elevated token when setting up otp

neurosnap:


grrowe:
It looks like it works intentionally however, on the Elevate login, I receive this warning/error. I can enter my 2FA code like usual, and it works.
image

joshuaj194 (approved):
Was able to successfully disable and enable 2FA without seeing the "access denied" error or the malformed QR code. I did see the same error when logging in that @grrowe pointed out. Could not use my yubikey but it worked with the 2FA code.

joshraker:
That error comes from the security key agent. You can't use a security key to log in from `localhost`. It's locked to our `dashbaord.aptible.com` and `app.aptible.com` domains. You have to use an OTP code.

PR #718, opened 2024-02-06
fix(endpoint): use same logic as `deploy-ui`

neurosnap:
https://aptible.slack.com/archives/C05CMUW3LHX/p1706829801698329

https://github.com/aptible/deploy-ui/blob/9ed4b5de1a3789e7fe1701050c5ba28a1662d6bd/app/models/vhost.js#L39

### DIFF
--- a/src/ui/shared/endpoint/util.tsx
+++ b/src/ui/shared/endpoint/util.tsx
@@ -80,6 +80,10 @@ export const EndpointStatusPill = ({
   );
 };
 
+const isWildcard = (url: string) => {
+  return url.startsWith("*");
+}
+
 export function EndpointUrl({ enp }: { enp: DeployEndpoint }) {
   const url = getEndpointUrl(enp);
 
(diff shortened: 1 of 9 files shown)


src/ui/shared/endpoint/util.tsx
```diff
@@ -80,24 +80,33 @@ export const EndpointStatusPill = ({
...
-  return (
-    <ExternalLink
-      className="text-black group-hover:text-indigo hover:text-indigo"
-      href={`https://${enp.virtualDomain}`}
-      variant="default"
-    >
-      https://{enp.virtualDomain}
-    </ExternalLink>
-  );
+  return <p>{url}</p>;
```
joshraker, line 109: Why did you choose a paragraph rather than fragment/span/div?

joshraker (approved):

src/ui/shared/endpoint/util.tsx
neurosnap: Good question.  I think it was what I saw already in this code so wanted to leave it as-is.

joshraker (approved):

PR #723, opened 2024-02-08
fix(endpoint): check container_ports as well

neurosnap:
https://aptible.slack.com/archives/C05CMUW3LHX/p1707366948501139


src/ui/pages/app-create-endpoint.tsx
```diff
@@ -85,7 +85,7 @@ export const AppCreateEndpointPage = () => {
   const [certId, setCertId] = useState("");
   const [privKey, setPrivKey] = useState("");
   const portText = getContainerPort(
-    { containerPort: port },
+    { containerPort: port, containerPorts: [] },
```
joshraker, line 88: Could get creative with the types so that we don't need to include the empty ports here.

joshraker (approved):

PR #726, opened 2024-02-08
fix: blank env editor

neurosnap:
Previously we showed all the environment variables which could cause some unintended side-effects and doesn't really map with how environment variables are set in Apps.  This change aligns more closely with how the CLI functions: it's additive.

I also updated the explainer text to make it more clear how we parse our textarea.
Screenshot 2024-02-08 at 7 52 25 PM

### DIFF
--- a/src/ui/pages/app-detail-config.tsx
+++ b/src/ui/pages/app-detail-config.tsx
@@ -1,13 +1,11 @@
 import {
   DeployCodeScanResponse,
   configEnvListToEnv,
-  configEnvToStr,
   createAppOperation,
   fetchApp,
   fetchConfiguration,
   hasDeployOperation,
   selectAppById,
-  selectAppConfigById,
 } from "@app/deploy";
 import {
   useDispatch,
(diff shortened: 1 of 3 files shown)


fancyremarker:
Tested functionality; matched expectations. 👍 

Added copy improvements:
image

Noticed we show this text on the page, which should only be relevant during FTUX. Did not remove so as not to break behavior elsewhere:

> Code scan information is only available for `git push` deployments.

fancyremarker:
Deferring to @almathew for reviewing and deciding on whether to take this approach.

almathew:
I'll 👍 once we've added Frank's updated language here https://aptible.slack.com/archives/C05CMUW3LHX/p1707434071436489?thread_ts=1707424541.084289&cid=C05CMUW3LHX

src/ui/pages/app-detail-config.tsx
```diff
@@ -232,7 +250,7 @@ export const AppDetailConfigPage = () => {
       <Box>
         <Group size="sm">
           <div className="flex items-center justify-between">
-            <h3 className={tokens.type.h3}>Edit Environment Variables</h3>
+            <h3 className={tokens.type.h3}>Add Environment Variables</h3>
```
almathew, line 253: This is a nit, but it's not really just Add, right? It can delete and overwrite too. I understand that Edit doesn't fully make sense, but it also does?

benjodo: Either Add/Modify/Delete or Edit make sense to me based on the context.

fancyremarker: You're right, reverted.

almathew (approved):

PR #728, opened 2024-02-09
fix(settings): use elevated token when revoking all tokens

neurosnap:
https://aptible.slack.com/archives/C05CMUW3LHX/p1707462063586959

### DIFF
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -81,7 +81,7 @@ export function setupTestStore(initialState: Partial<WebState>) {
   });
 
   store.run(function* (): Operation<void> {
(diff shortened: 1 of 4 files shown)


joshraker:
I tested this PR and it worked as expected.

1. Login to CLI with password
2. `aptible apps` works
3. Login to UI with SSO
4. Elevate into security settings
5. Revoke all tokens
6. `aptible apps` reports the token was revoked

I tested it on `main` to confirm I was able to reproduce the issue and was able to do so.

joshraker (approved):

joe-herman (approved):

PR #731, opened 2024-02-14
refactor(ftux): when to ask for a git push

neurosnap:
https://github.com/aptible/app-ui/assets/1940365/0cb1480e-0706-4d0e-aa2b-515c2af0fa43


joshraker (approved):
I feel like the final push/deploy instructions could be more clear/specific but not strongly enough to require it in this PR.

PR #734, opened 2024-02-21
chore: tighten spacing in FTUX header & tweak text

joeclayallday:
**Tighten spacing in FTUX header**

BEFORE

Screenshot 2024-02-21 at 5 15 12 PM

AFTER
Screenshot 2024-02-21 at 5 15 01 PM

---

Add clarity to Text

BEFORE
Screenshot 2024-02-21 at 5 48 50 PM

AFTER
Screenshot 2024-02-21 at 6 05 56 PM

---

Add Databases explainer text

BEFORE
Screenshot 2024-02-21 at 7 01 14 PM

AFTER
Screenshot 2024-02-21 at 7 00 14 PM

### DIFF
--- a/src/ui/pages/app-deploy-configure.tsx
+++ b/src/ui/pages/app-deploy-configure.tsx
@@ -227,8 +227,9 @@ export const AppDeployConfigurePage = () => {
 
         <FormGroup
           label="Databases"
+          htmlFor="databases"
           description="Aptible automatically handles configuring and managing databases for Elasticsearch, InfluxDB, MySQL, PostgreSQL, Redis and more."
-        ></FormGroup>
+        />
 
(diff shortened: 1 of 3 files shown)


src/ui/pages/app-deploy-configure.tsx
```diff
@@ -225,6 +225,12 @@ export const AppDeployConfigurePage = () => {
           )}
         </div>
 
+        <FormGroup
```
neurosnap, line 228: Hmm, I wonder if it would make more sense to recreate how this looks without using `FormGroup`?  My concern is that we are doing a bunch of stuff inside the component that isn't really applicable here.

neurosnap (approved):

PR #737, opened 2024-02-23
[DO NOT MERGE] Add Updated timeago styling to <Descbar> globally

joeclayallday:
**Notes**
• We can have it update every second, but I think the constant second counting will be distracting. I think a better solution is to change the "Updated N" text every minute (ex. 1m ago, 2m ago, etc...), and after 59m ago, change to 1h ago (1h ago, 2h ago, etc...).
• After refreshing change the "Updated N" text to **Updated Now**
• I think we can hide the refresh icon until the page hits 1m ago OR whatever time we think justifies a refresh

Normal State
Screenshot 2024-02-23 at 11 28 57 AM

Hover State
Screenshot-2024-02-23-at-11 29 00 AM

### DIFF
--- a/src/ui/shared/activity.tsx
+++ b/src/ui/shared/activity.tsx
@@ -263,7 +263,7 @@ function ActivityTable({
           </Group>
 
           <Group variant="horizontal" size="lg" className="items-center mt-1">
-            <DescBar>{paginated.totalItems} Operations</DescBar>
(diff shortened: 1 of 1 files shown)


joeclayallday:
We should add it to the Metrics pages too here:
Screenshot 2024-02-23 at 11 34 41 AM

neurosnap:
Closing for now until we can find time to work on it.

PR #739, opened 2024-02-27
feat(app): dependencies tab

neurosnap:


joeclayallday (approved):

src/deploy/configuration/selectors.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      };
+
+      const actual = selectDepGraphDatabases(state, { id: "app-id" }).map(
+        (db) => db.id,
+      );
+      expect(actual).toEqual(["1234", "5555"]);
+    });
+  });
+
+  describe("aptible-angel-studios.com", () => {
```
joshraker, line 120: Nit: Only one additional domain is necessary to confirm that it's using the stack's domain and not a hardcoded value.

joshraker (commented):
Code looks good. Need to pull it down and see how it works.

joshraker (approved):
This is quite neat. I'm excited to extend this to be able to save detected dependencies as formal dependencies once that model exists.

PR #755, opened 2024-03-07
Check for existing invitation during signup

joshuaj194:
Prevent a new user from signing up and accidentally creating a duplicate org when they have an existing invitation. Instead, show an error message that points them to check their email for an invite.

### DIFF
--- a/src/auth/signup.ts
+++ b/src/auth/signup.ts
@@ -15,7 +15,13 @@ const log = createLog("signup");
 export const signup = thunks.create<CreateUserForm>(
   "signup",
   function* onSignup(ctx, next) {
-    const { company: orgName, name, email, password, challenge_token } = ctx.payload;
+    const {
+      company: orgName,
+      name,
+      email,
+      password,
+      challenge_token
+    } = ctx.payload;
     const id = ctx.key;
     yield* schema.update(schema.loaders.start({ id }));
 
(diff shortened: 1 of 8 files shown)


joshuaj194:
@grrowe @neurosnap, the review on this needs to wait until the updates on Auth API are released (should happen tomorrow). @benjodo will update here once that is released

joshuaj194:
To test:
- send an invite to add a new user
- instead of using the link in the email, go straight to the signup page and try to sign up using the email address you sent the invite to.

The form should not submit. Instead, you should see an error telling you to check your email for an invite from support@aptible.com

src/auth/signup.ts
```diff
@@ -15,10 +15,35 @@ const log = createLog("signup");
 export const signup = thunks.create<CreateUserForm>(
   "signup",
   function* onSignup(ctx, next) {
-    const { company: orgName, name, email, password } = ctx.payload;
+    const {
+      company: orgName,
+      name,
+      email,
+      password,
+      challenge_token,
```
neurosnap, line 23: This should be `camelCase`

neurosnap (commented):
Overall looks good

joshuaj194 (commented):
updated to use camelCase for `challengeToken`

grrowe (approved):
Lgtm, manually tested.

PR #756, opened 2024-03-08
feat(app): deployments tab

neurosnap:


src/ui/shared/deployments-table.tsx
```diff
@@ -0,0 +1,162 @@
...
+        <Td>Tag</Td>
+        <Td>Commit</Td>
+        <Td>Message</Td>
+        <Td>Date</Td>
+        <Td variant="right">Actions</Td>
+      </THead>
+
+      <TBody>
+        {deployments.length === 0 ? (
+          <EmptyTr colSpan={7}>No deployments found</EmptyTr>
```
eabruzzese, line 152: I wonder if we should add a link to the deployment docs or something here?

src/ui/pages/app-detail-deployments.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <TestProvider>
+        <AppDetailDeploymentsPage />
+      </TestProvider>,
+    );
+
+    const links = await screen.findAllByRole("link", { name: "a947a95" });
+    expect(links[0].getAttribute("href")).toEqual(
+      "https://github.com/aptible/app-ui/commit/a947a95a92e7a7a4db7fe01c28346281c128b859",
+    );
+    // expect(screen.queryByText(/quay.io\/aptible\/cloud-ui/)).toBeInTheDocument();
```
eabruzzese, line 28: Still want this?

src/ui/shared/deployments-table.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  return (
+    <Tr>
+      <Td>
+        <Link
+          to={deploymentDetailUrl(deployment.id)}
+          className={tokens.type["table link"]}
+        >
+          {deployment.id}
+        </Link>
```
eabruzzese, line 101: I wonder if we should use the timestamp instead of the ID as the clickable thing? IDs can be short and harder to click.

src/ui/layouts/deployment-detail-layout.tsx
```diff
@@ -0,0 +1,130 @@
...
+          <DeploymentGitSha deployment={deployment} />
+        </DetailInfoItem>
+
+        <DetailInfoItem title="Docker Image">
+          <CopyText text={`${deployment.dockerImage}`} />
+        </DetailInfoItem>
+
+        <DetailInfoItem title="User">
+          {op.userName}
+          <div className="text-gray-500">Note: {op.note || "N/A"}</div>
```
eabruzzese, line 88: Should we just hide the note if there isn't one?

Dockerfile
```diff
@@ -1,11 +1,11 @@
...
-ARG VITE_AUTH_URL=https://auth.aptible.com
-ARG VITE_BILLING_URL=https://goldenboy.aptible.com
-ARG VITE_API_URL=https://api.aptible.com
+ARG VITE_APP_URL=https://app-sbx-eabruzzese.aptible-sandbox.com
+ARG VITE_AUTH_URL=https://app-68953.on-aptible.com
+ARG VITE_BILLING_URL=https://goldenboy-sbx-eabruzzese.aptible-sandbox.com
+ARG VITE_API_URL=https://app-68952.on-aptible.com
 ARG VITE_LEGACY_DASHBOARD_URL=https://dashboard.aptible.com
-ARG VITE_METRIC_TUNNEL_URL=https://metrictunnel.aptible.com
+ARG VITE_METRIC_TUNNEL_URL=https://metrictunnel-sbx-eabruzzese.aptible-sandbox.com
```
eabruzzese, line 8: Just a note to revert these before we merge

src/deployment/index.ts
```diff
@@ -0,0 +1,137 @@
...
+  },
+);
+
+export function getRegistryParts(url: string): { name: string; tag: string } {
+  const [name, tag] = url.split(":");
+  return { name, tag };
+}
+
+export function getRepoNameFromUrl(url: string): string {
+  return url.replace("https://", "");
```
eabruzzese, line 102: Maybe a little more robust to use a `URL` object here? Something like:

```suggestion
  const parsed = new URL(url)
  
  return [parsed.host, parsed.pathname.replace(/\.git$/, '')].join("")
```

That would give you:

```javascript
>>> const parsed = new URL("https://github.com/whatever/repo.git")
>>> [parsed.host, parsed.pathname.replace(/\.git$/, '')].join("")
'github.com/whatever/repo'

File: src/deployment/index.ts [hunk 7]
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  yield* next();
+
+  if (!ctx.json.ok) {
+    return;
+  }
+
+  ctx.loader = { message: "Rollback initiated" };
+});
```
eabruzzese, line 129: Still want this?

src/schema/factory.ts
```diff
@@ -798,3 +800,25 @@ export const defaultMembership = (m: Partial<Membership> = {}): Membership => {
...
+  return {
+    id: "",
+    dockerImage: "",
+    dockerRepositoryUrl: "",
+    status: "",
+    gitRepositoryUrl: "",
+    gitRef: "",
+    gitCommitSha: "",
+    gitCommitUrl: "",
+    gitCommitMessage: "",
```
eabruzzese, line 815: Do you want the commit date in here?

eabruzzese (commented):
Looking good!

src/ui/shared/deployments-table.tsx
neurosnap: This is consistent with how we render tables elsewhere, I'd prefer to do those all in a separate PR if we want to change it.

src/ui/layouts/deployment-detail-layout.tsx
neurosnap: This is how we have done it for awhile -- i just moved some stuff around -- so I'd prefer to leave it as-is for now.

src/deployment/index.ts
eabruzzese: Yeah that's basically the same thing. I was just removing the `.git` from the end to clean it up a bit, but we don't have to do that.

eabruzzese (approved):

PR #757, opened 2024-03-08
fix(elevate): only permit local redirects

neurosnap:


eabruzzese (approved):

src/ui/pages/elevate.tsx
```diff
@@ -57,7 +57,11 @@ export const ElevatePage = () => {
   const webauthnLoader = useLoader(webauthnAction);
 
   useLoaderSuccess(loader, () => {
-    navigate(redirect || homeUrl());
+    if (/^\/\w+/.test(redirect)) {
```
joe-herman, line 60: Silly question: how is this restricting to local redirects?

neurosnap: I still need to test it, but it will only accept `redirect` values with `/asd` *not* `https://asd.com` or even `//asd.com`

joe-herman: Ah okay that's simple enough, thanks

src/string-utils/index.ts
```diff
@@ -183,3 +183,8 @@ export const parseText = <
 
   return items;
 };
+
+export const allowLocalUrls = (url: string) => {
+  if (url === "/") return true;
+  return /^\/\w./.test(url);
```
eabruzzese, line 189: I think this technically works for this purpose, but did you mean for the pattern to only test the first character after the slash?

```javascript
> '/apps'.match(/^\/\w./)
['/ap', index: 0, input: '/apps', groups: undefined]
```

I think you may have wanted something more like this (which would also catch the root URL case, and support fragments if we wanted)?

```suggestion
return /^\/(.*)/.test(url);
```

eabruzzese (commented):
One small nit and a question about the intention of the regex, but it otherwise looks good!

src/string-utils/index.ts
```diff
@@ -183,3 +183,8 @@ export const parseText = <
 
   return items;
 };
+
+export const allowLocalUrls = (url: string) => {
```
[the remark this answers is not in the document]
neurosnap: Great!

src/string-utils/index.ts
neurosnap: Good catch!  I actually just wanted `\w+`

eabruzzese: @neurosnap I think using `\w+` there would break for paths with multiple slashes, e.g. `/apps/69875/services`, as well as anything that includes a query string, like `/environments?search=asdf`, right?

If you want to support those as well, we can either make the regex more complex, or just simplify this to check that the given string starts with a `/`.

neurosnap: I just want to make sure it starts with `/` and a character because I want to negate `//github.com` or w/e.  I added more test cases and they pass just fine as-is.

eabruzzese (approved):

PR #766, opened 2024-04-01
security: always use app 404

joe-herman:
## Summary
* security: route 404s to app

## Context
PCI compliance scanning also flagged that some 404 URLs returned the default nginx 404 pages. The behavior found was as follows:

https://app.aptible.com/foo returns an app-based 404, e.g.
image

https://app.aptible.com/foo.js returns an nginx 404, e.g.
image

We either need a custom nginx 404 page, or allow the app to handle 404s. This PR opts for the latter.

## Testing
Available on https://app-sbx-jherman.aptible-sandbox.com/

## References
* https://app.shortcut.com/aptible/story/23324

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -8,7 +8,6 @@ http {
   server {
     listen 80;
     server_name localhost;
-    server_tokens off;
     charset utf-8;
 
     root /app;
--- a/nginx.conf
+++ b/nginx.conf
@@ -8,13 +8,14 @@ http {
   server {
     listen 80;
     server_name localhost;
+    server_tokens off;
     charset utf-8;
 
     root /app;
     index index.html;
 
     location ~* \.(?:css|js)$ {
-        try_files $uri =404;
+        try_files $uri /index.html;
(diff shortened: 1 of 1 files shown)


nginx.conf
```diff
@@ -14,15 +14,15 @@ http {
...
+        try_files $uri /index.html;
         expires 1y;
         access_log off;
         add_header Cache-Control "public";
     }
 
     # Any route containing a file extension (e.g. /devicesfile.js)
     location ~ ^.+\..+$ {
-        try_files $uri =404;
+        try_files $uri /index.html;
```
mdelaossa, line 25: hah I was gonna complain that index.html doesn't actually route on a lot of modern apps but I see it has been cleverly chosen because rails actually knows what to do with it (show an app-based 404)... so I'm good with this

mdelaossa (approved):
LGTM. I'd prefer we had a global 404 to catch when an app can't deal, but on the worst case we're redirecting to root which isn't the end of the world

PR #768, opened 2024-04-02
Upgrade the aptible-deploy-action to v4

eabruzzese:


.github/workflows/deploy.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     needs: build
     steps:
       - uses: actions/checkout@v2
-      - uses: aptible/aptible-deploy-action@v1
+      - uses: aptible/aptible-deploy-action@v4
```
neurosnap, line 43: Should we also add `docker_repository_url`?

eabruzzese: Ah yeah, good call. We should probably also just derive it ourselves in the API for major providers.

neurosnap (approved):

PR #778, opened 2024-04-05
refactor(onboarding): send new users to hosting page

neurosnap:


src/app/router.tsx
```diff
@@ -511,7 +512,11 @@ export const appRoutes: RouteObject[] = [
         children: [
           {
             index: true,
-            element: <EnvironmentsPage />,
+            element: (
+              <NewUserToOnboarding>
```
neurosnap, line 516: Based on our conversations, I think it actually makes more sense to lift this to the root container so any authorized route will redirect, right?  Because otherwise if a user lands on this page and then navigates away immediately before we can fetch environment, it will not redirect to the hosting page.

joshraker: I'm fine with that risk seeing as this isn't a hard block. All they have to do to get past is click a button and I don't want to detract from their initial experience by risking them getting redirected if they reload in FTUX after clicking "Deploy Now".

The main reason I wanted it on `/environments` is because it's the default page and I wanted to catch someone who may have left the site then came back to the home page before choosing to "Deploy Now".

Additionally, expect the API will return quickly if they don't have any environments.

eabruzzese: In that case, wouldn't we want this at the very top level, i.e. in the `AllRequired` component?

joshraker: No, I don't want it to redirect them if they (re)load another page. Only if they navigate to the root URL. We can't put it on the home page at the root URL though because it will likely perform its logged-in check + redirect and be unmounted before it can perform the environment check. Hence, we put it on the page that the home page redirects to by default `/environments`.

eabruzzese (approved):

PR #783, opened 2024-04-09
fix: a bug preventing some users from viewing deployment configuration

eabruzzese:
This PR fixes a [reported bug](https://aptible.slack.com/archives/C05CMUW3LHX/p1712611841145119) that was attempting to fetch an app using a deployment ID (instead of the deployment's app ID), manifesting as a permission issue.


neurosnap (approved):
good find!

neurosnap (approved):

PR #787, opened 2024-04-10
feat: sort and filter apps by their deployment metadata on the source detail page

eabruzzese:
This PR implements support for searching and sorting apps on the Sources page based on their current git ref/commit, docker image, and deploy time.


neurosnap (commented):
nice work! just one comment

neurosnap (approved):

PR #790, opened 2024-04-12
Dockerfile dev

neurosnap:


eabruzzese (approved):
Approved!

FWIW, another option would be to adjust the regular `Dockerfile` to add a `dev` stage, and then use `services.app.build.target: dev` in the `docker-compose.yml`.

PR #791, opened 2024-04-12
feat(log-drain): show destination column

neurosnap:
refactor(log-drain): remove restart except for tails
refactor(metric-drain): remove restart
feat(metric-drain): show associated database


src/ui/pages/environment-detail-integrations.tsx
```diff
@@ -193,6 +198,75 @@ const LogDrainActions = ({ logDrain }: { logDrain: DeployLogDrain }) => {
   );
 };
 
+const isHttps: LogDrainType[] = [
```
eabruzzese, line 201: Maybe call this something like `HTTPS_LOG_DRAIN_TYPES` or something? `isHttps` reads like it's a function that returns a boolean, but it's actually a static list.

eabruzzese (approved):
Looks good! One small suggestion for readability, but approved.

PR #792, opened 2024-04-15
feat: source setup page

joeclayallday:
This PR introduces a "Source Setup" page. It also adds the Source URL to the source list table.

If a user has no sources, they will be automatically redirected to the Source Setup page, and given instructions on how to set up a source for a selected app.

Screenshot 2024-04-19 at 5 15 06 PM

A user that already has at least one source configured can visit the Source Setup page by clicking the New Source button.

Screenshot 2024-04-22 at 11 51 23 AM

### DIFF
--- a/src/ui/pages/sources-setup.tsx
+++ b/src/ui/pages/sources-setup.tsx
@@ -1,0 +1,88 @@
+import { selectAppsAsList } from "@app/deploy";
+import { useSelector } from "@app/react";
+import { appCiCdUrl } from "@app/routes";
+import { AppSidebarLayout } from "@app/ui";
+import {
+  Banner,
+  Box,
+  Button,
+  FormGroup,
+  Group,
+  Select,
+  SelectOption,
+  TitleBar,
+  tokens,
+} from "@app/ui/shared";
+import { useState } from "react";
+import { Link } from "react-router-dom";
+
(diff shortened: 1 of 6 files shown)


joeclayallday:
Tightened the spacing up a bit

Screenshot 2024-04-19 at 6 01 06 PM

src/ui/pages/sources-setup.tsx
```diff
@@ -0,0 +1,104 @@
...
+export function SourcesSetupPage() {
+  const { isLoading } = useQuery(fetchApps());
+  const apps = useSelector(selectAppsAsList);
+  const options: SelectOption[] = [
+    {
+      label: isLoading ? "Please wait..." : "Select an app",
+      value: "",
+      disabled: isLoading,
+    },
+    ...(apps || [])
```
neurosnap, line 28: `apps` should always be an empty list so this shouldn't be necessary.

neurosnap (commented):
1 comment, otherwise, looks good!

src/ui/pages/sources-setup.tsx
eabruzzese: Ah, this was to make TypeScript happy in a previous iteration, but it looks like it's no longer needed. Removed 👍

neurosnap (approved):

PR #795, opened 2024-04-15
fix: race between adding and removing memberships for user

neurosnap:
fix: reset edit membership form when success

https://aptible.slack.com/archives/C05CMUW3LHX/p1713206044643149

Previously we were submitting all changes to the API in parallel for add and removing memberships for a user.  Because these are all separate API requests, it is possible to run into a race condition where the memberships being removed complete before the requests being added complete.

When a user does not have any memberships for an organization, they are automatically removed.  This is the race issue that can happen.

The second issue happens when the user does multiple rounds of updating a user's roles on the same page.  When the changes are submitted and successfully, we we not properly "resetting" the form which caused weird issues like trying to create a membership that already exists.

Repo:
- User edits roles
- User saves
- Process succeeds
- User clicks save again
- Error! User is already has those roles

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "react-dom": "^18.2.0",
     "react-router": "^6.16.0",
     "react-router-dom": "^6.16.0",
-    "starfx": "^0.10.0",
+    "starfx": "^0.11.0",
     "tailwindcss": "^3.3.3",
     "typescript": "^5.2.2",
     "vite": "^4.5.3",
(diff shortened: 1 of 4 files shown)


eabruzzese (approved):
Approved, although I wonder if we should detect the situation where a user would be fully removed from the organization and prompt for confirmation?

neurosnap:
> Approved, although I wonder if we should detect the situation where a user would be fully removed from the organization and prompt for confirmation?

The UI blocks the user from being able to remove all roles for a user.  This was unfortunately a weird race condition because of the way we have to update a user's memberships (we have to make a request for each membership change).

PR #799, opened 2024-04-17
refactor: better tooltips

eabruzzese:
This PR refactors our `Tooltip` component to use [Floating UI](https://floating-ui.com/docs/react) for better performance and more robust handling of automatic placement.

For the most part, this change won't impact the user experience. This PR is laying groundwork for adding more context-specific help/info in the future. However, it does add a couple of tooltips for Sources and Deployments, mostly for testing while developing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   },
   "dependencies": {
     "@biomejs/biome": "^1.2.2",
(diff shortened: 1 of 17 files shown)


neurosnap (approved):
sweet!

PR #807, opened 2024-04-25
fix(env): change default values for env backup retention policy

neurosnap:
https://app.shortcut.com/aptible/story/24238/display-of-backup-retention-policy-on-a-new-environment-is-wrongly-set-to-0?vc_group_by=day&ct_workflow=all&cf_workflow=500000186


UserNotFound (approved):
Thanks!!!

PR #816, opened 2024-05-01
Implement a GitHub App integration

eabruzzese:


src/api/index.ts
```diff
@@ -310,6 +329,20 @@ metricTunnelApi.use(aborter);
...
+metricTunnelApi.use(debugMdw);
+metricTunnelApi.use(sentryErrorHandler);
+metricTunnelApi.use(expiredToken);
+metricTunnelApi.use(mdw.api({ schema }));
+metricTunnelApi.use(aborter);
+metricTunnelApi.use(portalApi.routes());
+metricTunnelApi.use(requestPortal);
+metricTunnelApi.use(aborter);
+metricTunnelApi.use(tokenMdw);
+metricTunnelApi.use(mdw.fetch());
```
neurosnap, line 344: `portalApi.use`

eabruzzese: 🤦 good lord how did I miss that. Thanks!

almathew (approved):

src/github-integration/index.ts
```diff
@@ -0,0 +1,120 @@
+import { cacheMinTimer, portalApi } from "@app/api";
+import { schema } from "@app/schema";
+import { GithubIntegration } from "@app/types";
+
+export const fetchGithubIntegrations = portalApi.get<
```
neurosnap, line 5: Does this endpoint ever return multiple integrations or can we assume it's always just a single record?  If we can make that assumption I think we could change the schema type to just a `slice.obj`

neurosnap (commented):
One tiny comment but otherwise looks good

src/github-integration/index.ts
eabruzzese: It does, yes. It's a less-common scenario for a company to have multiple Github accounts, but it happens:

* Acquisitions/mergers
* Migrating to/from GitHub enterprise
* Early startup shenanigans

It's also _slightly_ easier to implement this way on the backend, since we don't have to enforce that an organization can only have one GitHub integration.

In the future I could see this page evolving into more of a general "integrations" page, but this is good enough for now.

src/ui/pages/github-integration.tsx
```diff
@@ -0,0 +1,168 @@
...
+   * organization.
+   */
+  const [searchParams] = useSearchParams();
+  const installationId = searchParams.get("installation_id") || "";
+  const installAction = createGithubIntegration({ installationId });
+  const {
+    isLoading: isInstalling,
+    isError: isInstallError,
+    message: installError,
+  } = useLoader(installAction);
```
eabruzzese, line 55: @neurosnap do I need to do something in order to get `isError` working properly? At the moment it never becomes `true`, even though the endpoint returns a 4xx/5xx status.

Do I need to do something in the endpoint to update the state with an error somehow?

src/api/index.ts
```diff
@@ -310,6 +329,20 @@ metricTunnelApi.use(aborter);
...
+  createThunks({ supervisor: takeEvery }),
+);
+portalApi.use(debugMdw);
+portalApi.use(sentryErrorHandler);
+portalApi.use(expiredToken);
+portalApi.use(mdw.api({ schema }));
+portalApi.use(aborter);
+portalApi.use(portalApi.routes());
+portalApi.use(requestPortal);
+portalApi.use(aborter);
```
neurosnap, line 342: Remove this line

src/github-integration/index.ts
neurosnap: Sounds good, just wanted to make sure.

neurosnap (approved):

src/ui/pages/github-integration.tsx
eabruzzese: For posterity: this turned out to be a small bug in `starfx` that is now resolved.

neurosnap (approved):

src/ui/pages/github-integration.tsx
```diff
@@ -0,0 +1,168 @@
...
+            },
+            {
+              name: "GitHub Integration",
+              to: teamGithubIntegrationUrl(),
+            },
+          ]}
+        />
+
+        <FilterBar>
+          <div className="flex justify-end">
```
joeclayallday, line 79: @eabruzzese 👋 tiny tweak here I noticed, we can remove the justify-end and align the button left above the table. It removes the weird gap below the breadcrumb and people will see the button right away.

src/ui/pages/github-integration.tsx
```diff
@@ -0,0 +1,168 @@
...
+            <LoadingBar isLoading={isLoading} />
+          </Group>
+        </FilterBar>
+      </Group>
+
+      <Table>
+        <THead>
+          <Th>Account</Th>
+          <Th>Status</Th>
+          <Th variant="center">Actions</Th>
```
joeclayallday, line 98: @eabruzzese Another small note, use variant="right" to text-align right and pattern match all the other action column text formatting. Alright gonna log off now, nice work dude!

PR #821, opened 2024-05-03
VpnTunnel - show global and individual connection status

mdelaossa:
## When auto == "route"
image

## When auto has another value
image


src/deploy/vpn-tunnel/index.ts
```diff
@@ -72,9 +85,24 @@ export const deserializeDeployVpnTunnel = (
     createdAt: payload.created_at,
     updatedAt: payload.updated_at,
     stackId: extractIdFromLink(payload._links.stack),
+    state: payload.state as DeployVpnTunnelState,
+    tunnelAttributes: camelizeKeys(payload.tunnel_attributes),
   };
 };
 
+const camelizeKeys: any = (obj: any) => {
```
neurosnap, line 93: Can we move this into `string-utils`?

src/schema/factory.ts
```diff
@@ -574,6 +574,12 @@ export const defaultDeployVpnTunnel = (
     peerGateway: "",
     peerNetworks: [],
     perfectForwardSecrecy: "",
+    state: "unknown",
+    tunnelAttributes: {
+      connections: {},
```
neurosnap, line 579: What guarantees can we make about the fields on these objects?  Are they always present from the API?  Can they be null?  Where I'm going with this is we probably want to make guarantees here which means we should list out every field that needs to be available and provide sane defaults when we create this factory.

neurosnap (commented):
Overall looks good!  Nice job.  I have a couple of comments.

src/schema/factory.ts
mdelaossa: these fields can all be null (if present though they have all the attributes in the type), but it's guaranteed that they're the only three fields that would come in. Should I provide defaults as if these three fields along with their attributes came in, or s97ld I provide the minimal (a fully empty object?)

src/deploy/vpn-tunnel/index.ts
mdelaossa: it's not really a *string* utility, but sure!

neurosnap (approved):
nice job!

PR #830, opened 2024-05-21
Prototype: Add App dependencies to dependencies tab

joshraker:
Prototype for finding dependencies in App configurations without DNS resolution. The tests were updated to work but testing more examples e.g. matching vhost `externalHost` and `virtualDomain` would be good in future iterations.

Multiple config keys matching the same resource currently display as multiple rows.

Screenshot 2024-05-03 at 15 59 50


src/deploy/configuration/index.ts
```diff
@@ -96,63 +102,143 @@ export const selectAppConfigByAppId = createSelector(
...
+
   return deps;
 }
 
-export interface DepGraphDb extends DeployDatabase {
-  why: DepNode;
-}
+// export interface DepGraphDb extends DeployDatabase {
+//   why: DepNode;
+// }
```
joe-herman, line 204: not security related: still necessary?

src/deploy/configuration/index.ts
```diff
@@ -96,63 +102,143 @@ export const selectAppConfigByAppId = createSelector(
...
+    return depGroups;
   },
 );
 
-export interface DepGraphApp extends DeployApp {
-  why: DepNode;
-}
+// export interface DepGraphApp extends DeployApp {
+//   why: DepNode;
+// }
```
joe-herman, line 241: not security related: still necessary?

src/secrets/index.ts
```diff
@@ -0,0 +1,181 @@
+// Generated from https://github.com/gitleaks/gitleaks/blob/79cac73f7267f4a48f4bc73db11e105a6098a836/config/gitleaks.toml
```
joe-herman, line 1: question: if we wanted to regenerate this to get the latest set of regexes, how would we do that? Take `regex` items from the toml?

joe-herman (approved):
Thanks for adding the secrets check!

Nonblocking comments attached

src/secrets/index.ts
joshraker: "Derived" may be a more accurate description. I converted from TOML to JSON then it was whack-a-mole to address the JS unsupported regex features like `(?i)` (inline case insensitive flag).

joshraker: If we did want to update it, it'd probably be best to look at the toml diff from the linked commit and patch our regexes accordingly unless there's a lot of changes.

joshraker:
https://app.shortcut.com/aptible/story/24064/prototype-infer-connections-between-aptible-resources

neurosnap (approved):
nice job!

PR #855, opened 2024-06-05
Update deploy->Aptible, peer->remote

mdelaossa:


src/ui/pages/stack-detail-vpn-tunnels.tsx
```diff
@@ -136,11 +136,11 @@ export const StackDetailVpnTunnelsPage = () => {
             </TBody>
           </Table>
 
-          <p className="flex text-gray-500 text-base my-4">Deploy Networks</p>
+          <p className="flex text-gray-500 text-base my-4">Aptible Networks</p>
 
           <Table>
             <THead>
-              <Th>Network (As visible by peer)</Th>
+              <Th>Network (As seen from Remote)</Th>
```
fancyremarker, line 143: I think we should also update the language below (or leave this as "As visible by Remote" — I'm not wed to "As seen by")

Relatedly, is the language below incorrect? Should the Remote Networks list a "As routed by Remote" rather than "As routed by Aptible"?

See diff for what additional changes I'm suggesting: 

```diff
diff --git a/src/ui/pages/stack-detail-vpn-tunnels.tsx b/src/ui/pages/stack-detail-vpn-tunnels.tsx
index 00aa080c..053a8aa9 100644
--- a/src/ui/pages/stack-detail-vpn-tunnels.tsx
+++ b/src/ui/pages/stack-detail-vpn-tunnels.tsx
@@ -158,8 +158,8 @@ export const StackDetailVpnTunnelsPage = () => {
 
           <Table>
             <THead>
-              <Th>Network (As visible by Aptible)</Th>
-              <Th>Network (As routed by Aptible)</Th>
+              <Th>Network (As seen from Aptible)</Th>
+              <Th>Network (As routed by Remote)</Th>
             </THead>
 
             <TBody>
```

fancyremarker (changes requested):

src/ui/pages/stack-detail-vpn-tunnels.tsx
mdelaossa: Hmm, I think `as routed by Remote` like you wrote is probably more clear, since we can be given private IPs on that side that we do not control

fancyremarker (approved):
Thank you! I reviewed the current docs and there should be no change required there. (In the future, we may want to expand the docs, but I think they're fine for now.)

PR #859, opened 2024-06-18
fix(app): show deployment type and git remote

neurosnap:
Screenshot 2024-06-18 at 10 29 08 AM
Screenshot 2024-06-18 at 10 28 04 AM

### DIFF
--- a/src/ui/layouts/app-detail-layout.tsx
+++ b/src/ui/layouts/app-detail-layout.tsx
@@ -2,10 +2,12 @@ import { prettyDateTime } from "@app/date";
 import {
   cancelAppOpsPoll,
   fetchApp,
+  fetchConfiguration,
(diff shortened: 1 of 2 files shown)


src/ui/layouts/app-detail-layout.tsx
```diff
@@ -51,6 +55,12 @@ export function AppHeader({
   app,
   isLoading,
 }: { app: DeployApp; isLoading: boolean }) {
+  useQuery(fetchConfiguration({ id: app.currentConfigurationId }));
```
eabruzzese, line 58: Make sure to test this with a Basic Read user to make sure they don't get an error, since I think they don't have access to configs.

I think that's why we were putting all of the deployment details on the Deployment object.

PR #868, opened 2024-07-08
refactor: array.prototype.at => array[len-1]

neurosnap:
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at
Addresses: https://aptible.sentry.io/issues/5573197224/?alert_rule_id=14705789&alert_type=issue&notification_uuid=78c2294a-8285-469b-b2aa-a55da98d7264&project=4504990894391296&referrer=slack


joshraker (approved):
The `?.updatedAt || new Date().toISOString()` syntax would still work.

PR #882, opened 2024-07-12
Don't say 'all' if it's qualified with 'other'.

UserNotFound:
Be as clear as possible that "other" is not inclusive of the current session, including avoiding the word "all" entirely.

### DIFF
--- a/src/auth/token.ts
+++ b/src/auth/token.ts
@@ -170,9 +170,9 @@ export const revokeAllTokens = authApi.post(
     ctx.loader = {
       message:
         "Successfully logged out of other sessions! You may wish to log out of this session, as well.",
-     };
-   },
-  );
+    };
+  },
+);
 
 export function* revokeTokensMdw(ctx: AuthApiCtx, next: Next) {
   yield* next();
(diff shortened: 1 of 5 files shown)


neurosnap (approved):

src/auth/token.ts
```diff
@@ -167,7 +167,7 @@ export const revokeAllTokens = authApi.post(
       return;
     }
     tunaEvent("revoked-all-tokens");
-    ctx.loader = { message: "Successfully logged out all other sessions!" };
+    ctx.loader = { message: "Successfully logged out of other sessions! You may wish to log out of this session, as well." };
```
fancyremarker, line 170: Can we link to that page? https://app.aptible.com/logout

UserNotFound: I don't know typescript or this repo well enough, so unfortunately I don't know how to accomplish that. @neurosnap would you mind adding that?

UserNotFound: Screenshot 2024-07-15 at 08 38 56

neurosnap (approved):

PR #890, opened 2024-07-30
update database version + type to description

michaelwang13:
Switches from Type + Version to the Description, which has more information available that is already intended to be human readable (the same string you use when you _choose_ a database image)
Screenshot 2024-07-30 at 13 14 47
Screenshot 2024-07-30 at 13 15 24
Screenshot 2024-07-30 at 13 15 55


UserNotFound (approved):
Thanks!!!

PR #891, opened 2024-07-31
Don't create billing records in UI

aguilinger:
Requires https://github.com/aptible/auth-api/pull/593 before we can release this.

The linked auth change creates the records instead, so this would only throw errors. The errors don't block or anything, so the auth change can go first, but for keeping things clean, this would be nice to remove.


aguilinger:
https://app.shortcut.com/aptible/story/26339/investigate-abusive-trial-signup-oddities

neurosnap (approved):

PR #894, opened 2024-08-02
sort by hids start_date by default

michaelwang13:
https://aptible.zendesk.com/agent/tickets/54088

Customer reported that they use our HIDS reports for SOC2, and it's sometimes difficult to track because they're out of order. Updating logic to sort_by `From Date` instead of `Posted Date` by default. 

Before:
Screenshot 2024-08-02 at 6 32 37 PM

After: 

Screenshot 2024-08-02 at 6 33 25 PM

### DIFF
--- a/src/ui/pages/stack-detail-hids.tsx
+++ b/src/ui/pages/stack-detail-hids.tsx
@@ -124,7 +124,10 @@ const ReportTable = ({ stack }: { stack: DeployStack }) => {
     );
   }
   const lastPage = Math.ceil((data.total_count || 0) / (data.per_page || 0));
-  const reports = data._embedded.intrusion_detection_reports;
+  const reports = [...data._embedded.intrusion_detection_reports].sort(
+    (a, b) =>
+      new Date(b.starts_at).getTime() - new Date(a.starts_at).getTime(),
+  );
 
   return (
     <>
--- a/src/ui/pages/stack-detail-hids.tsx
+++ b/src/ui/pages/stack-detail-hids.tsx
@@ -76,9 +76,9 @@ const ReportView = ({
   const fdate = fileDate(report.created_at);
   return (
     <Tr>
-      <Td>{date}</Td>
       <Td>{prettyDate(report.starts_at)}</Td>
       <Td>{prettyDate(report.ends_at)}</Td>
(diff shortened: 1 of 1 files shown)


src/ui/pages/stack-detail-hids.tsx
```diff
@@ -124,7 +124,9 @@ const ReportTable = ({ stack }: { stack: DeployStack }) => {
     );
   }
   const lastPage = Math.ceil((data.total_count || 0) / (data.per_page || 0));
-  const reports = data._embedded.intrusion_detection_reports;
+  const reports = [...data._embedded.intrusion_detection_reports].sort(
+    (a, b) => new Date(b.starts_at).getTime() - new Date(a.starts_at).getTime(),
```
joe-herman, line 128: _Technically_ `.getTime()` is unnecessary (and maybe even creating a `new Date()`? But definitely still works with no credible performance hit

joe-herman (approved):
I'd like to petition to remove the Posted Date altogether

src/ui/pages/stack-detail-hids.tsx
michaelwang13: Just tested:
`b.starts_at` - `a.starts_at` doesn't work :(, and just `new Date()` isn't the right type for the comparison.

michaelwang13:
Thoughts on this @joe-herman? Swapping the Posted Date to the end. My typescript ability isn't great enough to make the table not look weird without the Posted Date column. 

Screenshot 2024-08-02 at 8 35 53 PM

src/ui/pages/stack-detail-hids.tsx
joe-herman: Oh interesting. I got a number when subtracting dates, thought it might carry over. Thanks for correcting me!

joe-herman (approved):

src/ui/pages/stack-detail-hids.tsx
michaelwang13: yeah, it seemingly doesn't work for all dates. (or it doesn't work at all and is defaulting to default Posted Date order) 
Screenshot 2024-08-02 at 8 47 53 PM

PR #901, opened 2024-08-07
SCIM

benjodo:
[sc-25542]

changes for adding/editing SCIM configuration

https://app.shortcut.com/aptible/story/25542/aptible-scim-implementation


src/ui/pages/settings-team-scim.tsx
```diff
@@ -0,0 +1,351 @@
...
+  </>
+);
+
+export const TeamScimPage = () => {
+  const scim = useCache(fetchScimConfigurations());
+
+  if (scim.isLoading) {
+    return <Loading />;
+  }
+  const configs = scim.data?._embedded?.scim_configurations || [];
```
joshraker, line 335: We usually use a selector to pull data that's fetched and stored by the various actions. However, I don't think I can fully describe what all needs changed in order to make that happen in the context of a PR and that seems excessive for just this one page so I'll leave it up to @neurosnap to decide if we want to get into that.

src/ui/shared/settings-sidebar.tsx
```diff
@@ -109,6 +112,23 @@ export function SettingsSidebar() {
...
+                Provisioning
+              </NavLink>
+            ) : (
+              <span className={navLink({ isActive: false })}>
+                Provisioning
+                <Tooltip text="Must be account owner" fluid>
+                  <IconLock variant="sm" className="ml-1 opacity-60" />
+                </Tooltip>
+              </span>
+            )}
```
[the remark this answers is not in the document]
benjodo: I'm open to feedback here, but my rationale in showing it disabled is that there may be a mix of users/roles in a given view which are SCIM-managed vs non-SCIM-managed, and I felt it was more useful for a user to see that they had the correct Aptible role to be able to admin those objects, but that those specific objects could not be mutated/destroyed b/c they were SCIM-managed.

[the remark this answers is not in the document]
benjodo: this has two components:
- whether or not to show the item, which is gated by feature flagging by org
- whether or not to link the item shown, which is limited to account owners, and follows establish patterns for other menu items like Single Sign-On

src/ui/pages/settings-team-members-edit.tsx
```diff
@@ -123,14 +126,20 @@ export function TeamMembersEditPage() {
...
                     name="roles"
-                    label={role.name}
+                    label={role.name + scimText}
                     key={role.id}
                     checked={userHasRole}
                     onChange={(e) => onChange(role.id, e.currentTarget.checked)}
-                    disabled={userHasRole && hasOnlyOneRole}
+                    disabled={
+                      (userHasRole && hasOnlyOneRole) ||
+                      (role.scimCreated && userIsScimManaged)
```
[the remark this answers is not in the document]
benjodo: that is a correct understanding of the behavior, and I think that's desired but I'll confirm with @apt-msteinlin

src/auth/scim.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const createScimConfiguration = authApi.post<CreateScimConfiguration>(
+  "/scim_configurations",
+  function* (ctx, next) {
+    ctx.request = ctx.req({
+      body: JSON.stringify({
+        scim_configuration: {
+          organization_id: ctx.payload.orgId,
+          default_role_id: ctx.payload.defaultRoleId,
+          unique_identifier: "email",
```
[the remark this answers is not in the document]
benjodo: I'll follow up with @apt-msteinlin on this one

src/ui/pages/settings-team-scim.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+const validators = {
+  roles: (data: Pick<CreateScimConfiguration, "defaultRoleId">) => {
+    if (!data.defaultRoleId) {
+      return "must select Default Role";
+    }
+  },
+};
+
+function ConfigureScim({ onSuccess }: { onSuccess: () => void }) {
```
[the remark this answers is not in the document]
benjodo: agreed, and I'll change this

src/ui/shared/settings-sidebar.tsx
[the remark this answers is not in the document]
joshraker: We're using the same pattern for other pages that are Account Owner only so this is 👍

src/schema/schema.ts
```diff
@@ -27,6 +27,7 @@ export const [schema, initialState] = createSchema({
   invitations: slice.table({ empty: factory.defaultInvitation() }),
   entities: slice.table<EmbeddedMap>(),
   redirectPath: slice.str(),
+  scimToken: slice.str(""),
```
neurosnap, line 30: `slice.str() == slice.str("")`

src/ui/pages/settings-team-scim.tsx
```diff
@@ -0,0 +1,351 @@
...
+    setShowTokenModal(false);
+    dispatch(resetScimToken());
+  };
+
+  const onSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    dispatch(action);
+  };
+
+  const onRemove = async () => {
```
neurosnap, line 172: no need for `async` here

src/ui/pages/settings-team-scim.tsx
neurosnap: One big goal of `starfx` is to allow [auto-caching](https://starfx.bower.sh/caching#automatic) because sometimes you really don't care about normalizing the data and you just need access to some API data in a single component.

So this is actually ideal to me.  However, if we start using the same endpoint in a bunch of components and need to do heavier CRUD operations then I would say we should upgrade this to a normal state slice.  Upgrading is also a goal for `starfx` so we should be good here.

neurosnap (commented):
wow! overall looks great!  let me know if you need anything else from me.  The only request would be to think about the minimum number of automated tests to write to feel confident we won't break anything in the future when we need to work on scim.  Even just a single test for any CRUD operations could be beneficial.

src/ui/pages/settings-team-members-edit.tsx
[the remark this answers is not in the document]
apt-msteinlin: @joshraker @benjodo  That is the correct behavior.

src/ui/pages/settings-team-scim.tsx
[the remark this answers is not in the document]
benjodo: (done)

src/ui/pages/role-detail-members.tsx
```diff
@@ -49,22 +48,67 @@ import {
...
 }: {
   membership: Membership;
   onDelete?: (id: string) => void;
   isLoading: boolean;
   canManage: boolean;
+  roleScimCreated: boolean;
 }) {
   const dispatch = useDispatch();
   const user = useSelector((s) => selectUserById(s, { id: membership.userId }));
+  const userIsScimManaged = !!user.externalId && user.externalId.trim() !== "";
```
neurosnap, line 106: I think this should probably be a selector since I see it duplicated a couple of times.

src/organizations/index.ts
```diff
@@ -108,7 +112,32 @@ export const selectHasBetaFeatures = createSelector(
     if (config.isDev) {
       return true;
     }
-    return ["df0ee681-9e02-4c28-8916-3b215d539b08"].includes(orgId);
+
+    // Array of organization IDs that have access to beta features
+    const betaFeatureOrgIds = config.betaFeatureOrgIds
+      .split(",")
+      .map((id) => id.trim())
+      .filter((id) => id);
```
neurosnap, line 120: `.filter(Boolean)`

src/ui/pages/settings-team-scim.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+            return res(
+              ctx.json({
+                _embedded: {
+                  scim_configurations: [{ ...testScim }],
+                },
+              }),
+            );
+          },
+        ),
+      );
```
neurosnap, line 32: I dont think you need this mock override since you are only testing for some text and not interacting with the page.

src/schema/factory.ts
```diff
@@ -58,6 +58,12 @@ export const defaultConfig = (e: Partial<Config> = {}): Config => {
     portalUrl: import.meta.env.VITE_PORTAL_URL || "",
     sentryDsn: import.meta.env.VITE_SENTRY_DSN || "",
     origin: (import.meta.env.VITE_ORIGIN as any) || "app",
+    betaFeatureOrgIds:
+      import.meta.env.VITE_FEATURE_BETA_ORG_IDS ||
+      "df0ee681-9e02-4c28-8916-3b215d539b08",
+    scimFeatureOrgIds:
+      import.meta.env.VITE_FEATURE_SCIM_ORG_IDS ||
+      "df0ee681-9e02-4c28-8916-3b215d539b08",
```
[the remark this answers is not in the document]
benjodo: I *think* so, because this is the Aptible org id, and it is what was in the codebase previously, but I'm open for feedback if you think we should remove a default value

src/auth/scim.ts
```diff
@@ -0,0 +1,139 @@
...
+  ctx.request = ctx.req({
+    body: JSON.stringify({
+      scim_configuration_id: ctx.payload.scimConfigurationId,
+      user_id: ctx.payload.userId,
+    }),
+  });
+  yield* next();
+
+  if (!ctx.json.ok) {
+    throw new Error("API request failed");
```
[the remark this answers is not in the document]
benjodo: thanks. so literally:
```
  if (!ctx.json.ok) {
    return ctx.json;
  }
```

src/ui/pages/settings-team-scim.test.tsx
benjodo: I tested with and without and it looks like I do, in order to pull in a scim configuration for the org, in the current setup

src/ui/pages/role-detail-members.tsx
benjodo: thanks; selector added.

src/ui/pages/settings-team-scim.tsx
```diff
@@ -0,0 +1,353 @@
...
+  <h2 className={tokens.type.h2}>Provisioning (SCIM)</h2>
+);
+
+const SharedScimInfo = () => (
+  <>
+    <Banner>
+      To configure System for Cross-domain Identity Management (SCIM), enter the
+      required information below. Terminology and acronyms vary between
+      providers. For Okta, follow our{" "}
+      <ExternalLink variant="default" href="https://www.aptible.com/docs/">
```
neurosnap, line 303: Does this link need updated?

src/ui/shared/settings-sidebar.tsx
```diff
@@ -109,6 +114,26 @@ export function SettingsSidebar() {
           </span>
         )}
 
+        {hasScimFeature && (
```
neurosnap, line 117: This is somewhat of a style thing but I would prefer we use a ternary here: `{hasScimFeature ? () : null}`

src/ui/pages/settings-team-scim.tsx
```diff
@@ -0,0 +1,353 @@
...
+                    options={options}
+                    onSelect={(opt) => setDefaultRoleId(opt.value)}
+                    value={defaultRoleId}
+                    ariaLabel="default-role-edit"
+                  />
+                </div>
+              </FormGroup>
+
+              <div>
+                <Button onClick={handleGenerateToken} isLoading={isLoading}>
```
neurosnap, line 231:
```
isLoading={generateTokenLoader.isLoading}
```

src/ui/pages/settings-team-scim.tsx
benjodo: yes, updated.

src/users/selectors.ts
```diff
@@ -64,3 +64,9 @@ export const selectIsUserVerified = createSelector(
   selectCurrentUser,
   (user) => user.verified,
 );
+
+export const selectIsUserScimManaged = createSelector(
+  (state: any, props: { id: string }) =>
+    selectUserById(state, { id: props.id }),
+  (user) => !!user?.externalId && user.externalId.trim() !== "",
```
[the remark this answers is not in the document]
benjodo: Thanks. I removed the conditional test for user (`user?`) but still need to test for both null and empty-string conditions on `user.externalId`

neurosnap (approved):
:shipit:

PR #922, opened 2024-08-21
update-environment-tooltip-text

joeclayallday:
**Update tooltip text to:** Environments in Stacks group resources and share networks with others on the same Stack.

Screenshot 2024-08-21 at 10 42 11 AM


joe-herman (approved):
That was quick

PR #923, opened 2024-08-22
add primary db link in header for replicas

benjodo:
https://app.shortcut.com/aptible/story/26534/the-cluster-tab-only-shows-replicas

show link in the db detail page header if the replica has a linked primary

image

image

### DIFF
--- a/src/ui/layouts/database-detail-layout.tsx
+++ b/src/ui/layouts/database-detail-layout.tsx
@@ -35,6 +35,7 @@ import {
 import { setResourceStats } from "@app/search";
 import type { DeployDatabase, DeployService } from "@app/types";
 import { useEffect, useMemo } from "react";
+import { Link } from "react-router-dom";
 import { Outlet, useParams } from "react-router-dom";
 import { usePoller } from "../hooks";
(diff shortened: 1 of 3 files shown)


joeclayallday (approved):

src/ui/layouts/database-detail-layout.tsx
```diff
@@ -105,6 +108,13 @@ export function DatabaseHeader({
         <DetailInfoItem title="Created">
           {prettyDateTime(database.createdAt)}
         </DetailInfoItem>
+        {databasePrimary?.id && (
```
neurosnap, line 111: With most of our selectors (like `selectDatabaseById`), we guarantee the object will exist with the correct properties.  So you don't need the `?.` in this case, just `.`.

neurosnap (changes requested):

src/ui/layouts/database-detail-layout.tsx
benjodo: accepted/changed 👍

neurosnap (approved):

PR #924, opened 2024-08-23
Improve backup load performance and show spinner until estimate is calculated

joshraker:
- I ran through all the places we show a cost estimate to ensure all of the necessary data is being pulled especially backups since they're not loaded on bootup.
- Moved fetching backups to per-database and per-environment in their lists so that we don't have to wait on all backups to get estimates. Stack estimates still require waiting for all backups to be fetched.
- Moved the cost loading spinner from headers to the cost themselves to make it extra clear when the cost has finished calculating (see screenshot below).

Screenshot 2024-08-23 at 16 37 48

### DIFF
--- a/src/ui/shared/db/database-list.tsx
+++ b/src/ui/shared/db/database-list.tsx
@@ -4,14 +4,17 @@ import {
   type DeployDatabaseRow,
   calcMetrics,
   estimateMonthlyCost,
-  fetchBackups,
+  fetchBackupsByDatabaseId,
   fetchBackupsByEnvId,
   fetchDatabaseImages,
   fetchDatabases,
   fetchDatabasesByEnvId,
+  fetchEndpoints,
+  fetchEndpointsByEnvironmentId,
   fetchEnvironmentById,
   fetchEnvironments,
   fetchServices,
+  selectBackupsByDatabaseId,
   selectDatabaseImageById,
   selectDatabasesForTableSearch,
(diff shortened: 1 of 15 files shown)


src/ui/shared/db/database-list.tsx
```diff
@@ -95,22 +98,41 @@ const DatabaseIdCell = ({ database }: DatabaseCellProps) => {
...
 };
 
-const DatabaseCostCell = ({ database }: DatabaseCellProps) => {
+const DatabaseCostCell = ({
+  database,
+  costLoading,
+  fetchBackups = false,
+}: DatabaseCellProps & { costLoading: boolean; fetchBackups?: boolean }) => {
+  let loading = costLoading;
+  if (fetchBackups) {
```
neurosnap, line 107: You can't have a hook inside of a conditional

neurosnap (changes requested):

src/ui/shared/db/database-list.tsx
joshraker: You can as long as the condition doesn't change. Or is that just the only time that react won't yell at you?

neurosnap: Yeah I would avoid it

neurosnap (approved):
Looks good.  I would say that this feels really complex for the FE to handle and we should push to do more of this work on the BE.

PR #935, opened 2024-09-12
Add Service settings area to configure ZDD

mdelaossa:
Depends on https://github.com/aptible/sweetness/pull/1771


src/ui/pages/app-detail-service-settings.tsx
```diff
@@ -0,0 +1,141 @@
...
+                      onChange={(e) =>
+                        setNextService({
+                          ...nextService,
+                          naiveHealthCheck: e.currentTarget.checked,
+                        })
+                      }
+                    />
+                  </Tooltip>
+                </div>
+                <div className="flex">
```
joeclayallday, line 114: Add `mt-4` here to add spacing between the buttons and the checkboxes

src/ui/layouts/service-detail-layout.tsx
```diff
@@ -64,6 +65,23 @@ export function ServiceHeader({
   const { totalCPU } = calcMetrics([service]);
   const [isOpen, setOpen] = useState(true);
 
+  const getDeploymentStrategy = () => {
+    if (endpoints.length > 0) {
+      return "Zero Downtime";
```
joeclayallday, line 70: We say Zero-Downtime with a hypen elsewhere. We should be consistent. So maybe add a hypen here OR remove it elsewhere.

joeclayallday (approved):

neurosnap (changes requested):
We don't need `modifyService` in this case so we can safely remove it.

neurosnap (approved):
nice job!

PR #964, opened 2024-11-04
chore: fix last scale banner

neurosnap:
We cannot always find the last (2) scale operations in order to render the from and to container profile and sizes.  This change makes it so we do not show inaccurate information to the user.

Further, scale operations do not always have all the properties we want to present to the user so we need to do some parsing to properly render the text.  Basically I tried to make sure we only show the user information if it is accurate.

Screenshot 2024-11-05 at 3 46 50 PM
Screenshot 2024-11-05 at 3 47 09 PM
Screenshot 2024-11-05 at 3 49 11 PM
Screenshot 2024-11-05 at 3 49 41 PM
Screenshot 2024-11-05 at 3 51 29 PM
Screenshot 2024-11-05 at 3 57 52 PM


src/deploy/operation/index.ts
```diff
@@ -309,41 +295,30 @@ export const selectNonFailedScaleOps = createSelector(
...
 
-    scaleAttrs.forEach((attr) => {
-      (current as any)[attr] = findOperationValue(lastOps, attr);
-    });
+    const [latest, prev] = ops.slice(0, 2);
+
+    // update previous operation based on what we can find in the op history
+    // ops is sorted in descending order based on updated_at so once we have
+    // a real value inside `prev` we dont update the field again
+    for (let i = 1; i < ops.length; i += 1) {
```
joshraker, line 310: Nit: You can start at `i = 2` because `ops[1]` is `prev` and comparing to itself won't yield any difference.

src/deploy/operation/index.ts
```diff
@@ -173,24 +170,13 @@ export const deserializeDeployOperation = (
...
   };
 };
 
-// Search an array of operations for the first that has the specified attribute
-// If none are found, use the value from the defaultDeployOperation
-export const findOperationValue = <K extends keyof DeployOperation>(
-  ops: DeployOperation[],
-  attr: K,
-) => {
-  const op = ops.find((op) => op[attr] != null);
```
joshraker, line 190: This is close enough to the solution we discussed with the product team that I won't let it hold up the PR but the current solution does not backfill the latest operation, only the previous operation. If we want to backfill both the only thing that needed to change from a few commits ago (though it may have been lost to the force push) is:

```typescript
const op = ops.find((op) => [null, -1, ""].includes(op[attr]));
```

This would account for the changes to the default container count, size, and instance profile.

joshraker (approved):

PR #979, opened 2024-11-20
add-stack-deprovision-tab

joeclayallday:
**Todo**
• Only show this Settings tab for dedicated stacks and self-hosted stacks


src/ui/layouts/stack-detail-layout.tsx
```diff
@@ -152,6 +153,10 @@ function StackPageHeader() {
     tabs.push({ name: "Managed HIDS", href: stackDetailHidsUrl(id) });
   }
 
+  if (stack.dedicatedOrSelfHostedStack) {
```
joeclayallday, line 156: Missing logic

PR #984, opened 2024-11-25
refactor: use cost estimates api

neurosnap:
https://app.shortcut.com/aptible/story/29248/use-new-costs-api-in-the-ui?vc_group_by=day&ct_workflow=all&cf_workflow=500000186

### DIFF
--- a/src/deploy/cost/calc.ts
+++ b/src/deploy/cost/calc.ts
@@ -6,10 +6,9 @@ import type {
   DeployStack,
   InstanceClass,
 } from "@app/types";
-import { CONTAINER_PROFILES } from "./container";
+import { CONTAINER_PROFILES } from "../container";
 
-export const hoursPerMonth = 731;
-
+export const hoursPerMonth = 730;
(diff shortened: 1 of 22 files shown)


src/deploy/cost/calc.ts
```diff
@@ -6,10 +6,9 @@ import type {
   DeployStack,
   InstanceClass,
 } from "@app/types";
-import { CONTAINER_PROFILES } from "./container";
-
-export const hoursPerMonth = 731;
+import { CONTAINER_PROFILES } from "../container";
 
+export const hoursPerMonth = 730;
```
mdelaossa, line 11: 730 vs 731?

mdelaossa (approved):

src/deploy/cost/calc.ts
neurosnap: It's the magic number we use on the BE: https://github.com/aptible/deploy-api/blob/main/app/lib/customer_metrics.rb#L13

mdelaossa (approved):

PR #993, opened 2024-12-04
token_header_implementation

michaelwang13:
Exposes the token header field for modification/use during app endpoint creation, details, and settings. This is feature flagged by org, and will be made available to CFA through env variable.


michaelwang13:
Need this: https://github.com/aptible/sweetness/pull/1816 deployed.

benjodo:
holding for https://github.com/aptible/sweetness/pull/1816 deployment.

benjodo (approved):

PR #1000, opened 2024-12-11
Operation filters: Update visible Operation types

mdelaossa:
Updated the filter list to only show types as of https://www.aptible.com/docs/core-concepts/architecture/operations, hiding internal ones

image


eloisejones (approved):
Verified against provided list

PR #1006, opened 2024-12-17
search-results-cursor-tweak

joeclayallday:
Change cursor pointer behavior for search results list

https://github.com/user-attachments/assets/53cb6e20-8a10-4c57-b359-5ea58cf9885e


UserNotFound (approved):
:that-was-easy:

PR #1015, opened 2024-12-30
Configure source maps for Sentry

eabruzzese:
This PR configures [source map uploads for Sentry + Vite](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/vite/).


.env.example
```diff
@@ -5,3 +5,8 @@ VITE_BILLING_URL="https://goldenboy.aptible.com"
 VITE_LEGACY_DASHBOARD_URL="https://dashboard.aptible.com"
 VITE_METRIC_TUNNEL_URL="https://metrictunnel-nextgen.aptible.com"
 VITE_PORTAL_URL="https://portal.aptible.com"
+
+# Leave the auth token unset unless you're testing Sentry integration
+SENTRY_AUTH_TOKEN=
+SENTRY_ORG=aptible
+SENTRY_PROJECT=app-ui-sbx-main
```
eabruzzese, line 12: Note that these are not prefixed with `VITE_` like the others, because we don't want their values exposed in the vite bundle.

almathew (approved):

PR #1038, opened 2025-02-03
fix: ability to filter by Service in App or Database activity pages

neurosnap:
https://aptible.slack.com/archives/C05CMUW3LHX/p1738592209907599

### DIFF
--- a/src/deploy/operation/index.ts
+++ b/src/deploy/operation/index.ts
@@ -511,7 +511,7 @@ export const fetchOperationsByAppId = api.get<
   { id: string } & PaginateProps & FilterOpProps,
   HalEmbedded<{ operations: DeployOperationResponse[] }>
 >(
-  "/apps/:id/operations?page=:page",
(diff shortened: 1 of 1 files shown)


joshuaj194:
This fixed the Activity tab within Apps but not within Environments where it still shows 0 operations when filtering for App and Scale:

image

joshuaj194:
works as expected in App Activity

image

joshuaj194 (changes requested):
See comments and screenshots above

neurosnap:
> This fixed the Activity tab within Apps but not within Environments where it still shows 0 operations when filtering for App and Scale

So this has not been baked into the API and I'm not fully convinced we want this behavior.  We can create a separate shortcut story for that issue specifically since it will involve more work to coordinate feature work.

My gut is telling me we should merge/deploy as-is and then have a separate shortcut story to tackle filtering by App->Scale.  Thoughts?

https://app.shortcut.com/aptible/story/30905

joshuaj194 (approved):

PR #1048, opened 2025-02-11
Update autoscaling UI wording and defaults for post release cooldown

aguilinger:
This won't actually go out until the deploy-api side is released, but in preparation, we want to make sure the interfaces have updated language and defaults.

https://github.com/aptible/deploy-api/pull/1684 and https://aptible.slack.com/archives/C022BUSF0B1/p1738846349350639 for full context.

### DIFF
--- a/src/ui/pages/app-detail-service-scale.tsx
+++ b/src/ui/pages/app-detail-service-scale.tsx
@@ -793,7 +793,7 @@ const AutoscalingSection = ({
                   </FormGroup>
                   <FormGroup
                     splitWidthInputs
-                    description="The number of seconds to wait after a general scale to potentially take another action"
+                    description="The number of seconds ignore in metrics after a scale event to allow for service stabilization"
                     label="Post Release Cooldown"
                     htmlFor="release-cooldown"
                   >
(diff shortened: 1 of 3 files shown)


src/ui/pages/app-detail-service-scale.tsx
```diff
@@ -793,7 +793,7 @@ const AutoscalingSection = ({
                   </FormGroup>
                   <FormGroup
                     splitWidthInputs
-                    description="The number of seconds to wait after a general scale to potentially take another action"
+                    description="The number of seconds ignore in metrics after a scale event to allow for service stabilization"
```
neurosnap, line 796: Reading this feels a little awkward, thoughts on: `The number of seconds to ignore after a scale event in order to allow for service stabilization`

eloisejones (approved):
🚗

src/ui/pages/app-detail-service-scale.tsx
aguilinger: That's what I had before elsewhere but there was feedback to add the 'in metrics'

PR #1050, opened 2025-02-11
fix: [sc-31180] app-ui: restricted pages cannot be accessed directly via URL

neurosnap:
Story details: https://app.shortcut.com/aptible/story/31180


joshraker (approved):
Nice, I was hoping we had something separate from `onFetchResourceData` for required data just like this.

PR #1054, opened 2025-02-20
Add support for displaying correlated anomalies and summaries for diagnostic dashboards

eabruzzese:
This PR adds support for displaying "ranked" plots (e.g., those that Hotshot found interesting and potentially correlated) and summary text.

> ⚠️ This PR must accompany https://github.com/aptible/hotshot/pull/12 when deployed.

Screenshot 2025-02-20 at 17 36 22@2x

### DIFF
--- a/src/ui/hooks/use-dashboard.ts
+++ b/src/ui/hooks/use-dashboard.ts
@@ -1,4 +1,4 @@
-import type { Message, Resource } from "@app/aptible-ai";
+import type { Message, Plot, Resource } from "@app/aptible-ai";
 import { selectAptibleAiUrl } from "@app/config";
 import { useSelector } from "@app/react";
 import { selectAccessToken } from "@app/token";
@@ -10,6 +10,8 @@ type Dashboard = {
     [key: string]: Resource;
   };
   messages: Message[];
+  summary: string;
+  ranked_plots: Plot[];
 };
 
 type UseDashboardParams = {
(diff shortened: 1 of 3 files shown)


src/ui/hooks/use-dashboard.ts
```diff
@@ -140,6 +150,9 @@ export const useDashboard = ({
         start_time: startTime,
         end_time: endTime,
       },
+      heartbeat: {
+        timeout: 60000 * 30, // 30 minutes
+      },
```
eabruzzese, line 155: This is a bugfix for [#31326](https://app.shortcut.com/aptible/story/31326/the-hotshot-websocket-disconnects-before-investigations-are-complete-if-they-take-too-long)

src/ui/shared/diagnostics/resource.tsx
```diff
@@ -81,7 +81,7 @@ export const DiagnosticsResource = ({
   endTime: string;
   synchronizedHoverContext: React.Context<HoverState>;
 }) => {
-  const [isCollapsed, setIsCollapsed] = useState(false);
+  const [isCollapsed, setIsCollapsed] = useState(true);
```
eabruzzese, line 84: Collapsing these by default because too many charts can crash the browser tab. We'll need to do some optimization there with something like [`react-intersection-observer`](https://github.com/researchgate/react-intersection-observer) to unmount components when they're not in the viewport.

almathew (approved):

PR #1068, opened 2025-03-25
Add Sorting to Columns in Stack Overview

benjodo:
# Add Sorting to Stacks Page

This PR adds sorting functionality to the Stacks page, modeled after the Environments page implementation.

Internal Enhancement Request: https://aptible.slack.com/archives/C05EBR77YRE/p1741795693225349

## Changes
- Added sortable columns for Name, ID, Region, Environments, Apps, Databases, and Cost
- Implemented sorting UI components matching the Environments page style
- Refactored stack table sorting logic to match the idiomatic approach used in the environment table
- Enabled Limits and Stack Type columns remain unsorted

## Testing
- Verify sorting works for all columns in both ascending and descending order
- Confirm sorting UI matches the Environments page (chevron icons, hover states)
- Test sorting behavior with search filters applied

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "react-dom": "^18.3.1",
     "react-router": "^7.0.2",
     "react-router-dom": "^7.0.2",
-    "react-use-websocket": "^4.13.0",
+    "react-use-websocket": "^4.11.1",
     "starfx": "^0.13.4",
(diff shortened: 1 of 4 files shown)


eabruzzese (approved):
Looks great! Tried it locally and it works well 👍

PR #1077, opened 2025-04-03
Remove backup retention policy announcement banner

benjodo:
This PR removes the backup retention policy announcement banner that was added for the July 2024 policy change ([PR 998](https://github.com/aptible/app-ui/pull/998)). The banner has served its purpose of informing users but currently reappears for users with cleared cookies, which is no longer a desired behavior.

Changes:
- Remove the notice component and related code
- Clean up schema and types
- Remove notice from app layout

Testing:
- [x] clear cookies, observe that the banner shows in prod 
- [x] clear cookies, observe that the banner does not show in local dev

image


eloisejones (commented):
If CS/CX is ready to get rid of this notice, I'm game. My main concern here is: 

Are we sure we want to scrape out all of the notice functionality? Wouldn't we want to leave some amount of it in place for the next time we need a notice?

I would be inclined to scrape out anything particular to the backup retention policy, but leave the scaffolding in place for the next notice.

joshraker:
> If CS/CX is ready to get rid of this notice, I'm game. My main concern here is:
> 
> Are we sure we want to scrape out all of the notice functionality? Wouldn't we want to leave some amount of it in place for the next time we need a notice?
> 
> I would be inclined to scrape out anything particular to the backup retention policy, but leave the scaffolding in place for the next notice.

This

eloisejones (approved):
Chatted with Ben about this offline. Good to proceed.

PR #1088, opened 2025-04-16
make container ports editable for tcp/tls endpoints

michaelwang13:
[sc-32570]

This PR allows `container_ports` to be edited for TCP and TLS endpoints. The current iteration of the endpoint edits forms only changes `container_port,` which is not relevant for TCP and TLS endpoints and hits a validation error. 

It also changes the port validation from using `Number.parseInt` to just `Number,` since the former has weird behavior — for example, `8asdfsadfsaf` would be considered `8` by the validator.

### DIFF
--- a/src/ui/pages/endpoint-detail-settings.tsx
+++ b/src/ui/pages/endpoint-detail-settings.tsx
@@ -4,6 +4,7 @@ import {
   getContainerPort,
   isRequiresCert,
   parseIpStr,
+  parsePortsStrToNum,
   selectAppById,
   selectEndpointById,
   selectImageById,
@@ -37,7 +38,22 @@ import {
 } from "../shared";
 
 const validators = {
-  port: (data: EndpointUpdateProps) => portValidator(data.containerPort),
+  port: (data: EndpointUpdateProps) => {
+    if (data.enpType !== "tls" && data.enpType !== "tcp") {
+      return portValidator(data.containerPort);
+    }
+  },
+  ports: (data: EndpointUpdateProps) => {
+    if (data.enpType === "tls" || data.enpType === "tcp") {
+      const errs: string[] = [];
+      data.containerPorts.forEach((port) => {
+        const result = portValidator(port.toString());
+        if (result) errs.push(`[${port}] is not between 1 and 65535`);
+      });
+      if (errs.length === 0) return;
+      return errs.join(", ");
+    }
+  },
   ipAllowlist: (data: EndpointUpdateProps) => ipValidator(data.ipAllowlist),
   cert: (data: EndpointUpdateProps) => {
     if (data.requiresCert && data.certId === "" && data.cert === "") {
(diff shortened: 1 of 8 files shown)


src/ui/pages/endpoint-detail-settings.tsx
```diff
@@ -201,22 +224,42 @@ const EndpointSettings = ({ endpointId }: { endpointId: string }) => {
...
+        description={`Current container ports: ${curPortText}`}
+        htmlFor="ports"
+        feedbackMessage={errors.ports}
+        feedbackVariant={errors.ports ? "danger" : "info"}
+      >
+        <Input
+          type="text"
+          id="ports"
+          name="ports"
+          value={ports}
```
fancyremarker, line 242: This is rendering with an `s` in between ports instead of a space or a comma currently. Example is Vhost 106218:
image

src/ui/pages/endpoint-detail-settings.tsx
```diff
@@ -201,22 +224,42 @@ const EndpointSettings = ({ endpointId }: { endpointId: string }) => {
...
-        type="text"
-        id="port"
-        name="port"
-        value={port}
-        onChange={(e) => setPort(e.currentTarget.value)}
-      />
-    </FormGroup>
+    isTlsOrTcp ? (
+      <FormGroup
+        label="Container Ports"
```
fancyremarker, line 232: Good to pluralize here. We probably also want to say "Ports" or "Container Ports" on the layout detail page (endpoint-detail-layout.tsx), for TCP endpoints where there can be multiple:

image

src/ui/pages/endpoint-detail-settings.tsx
fancyremarker: Also, weirdly, this text doesn't change even when `container_ports` is updated? Here's after (successfully) modifying the ports to be just 7835:
image

fancyremarker (changes requested):
In addition to the requested changes, and more importantly, this appears to break updating the `container_port` for HTTPS endpoints. For example, when trying to update the port for Vhost 106508 (type `http_proxy_protocol`) from default (80) to 443, the API request to `PATCH /vhosts/106508` includes the payload:

```json
{
  "container_port": "443",
  "container_ports": [80]
}
```

which results in a 422 error with response:

```json
{
    "code": 422,
    "error": "unprocessable_entity",
    "message": "Validation failed: Container ports must be blank for a HTTPS Endpoint"
}
```

src/ui/pages/endpoint-detail-settings.tsx
michaelwang13: Was missing the `\` in my join, but turns out it doesn't seem like you can join an array on numbers like that, so replaced with `, ` instead. 

The text stay was because I was incorrectly grabbing `exposedPorts` instead of looking at container ports on line 85. This also fixes the problem of breaking HTTP(s) endpoints.

src/ui/pages/endpoint-detail-settings.tsx
michaelwang13: Port for HTTP
Screenshot 2025-04-17 at 6 10 54 PM

Ports for TLS/TCP
Screenshot 2025-04-17 at 6 12 41 PM

src/ui/pages/endpoint-detail-settings.tsx
michaelwang13: Example, when swapping the port on a HTTP endpoint from the default 8000 to 443, it no longer includes 8000 in container_ports in the payload: 
```
{
   container_port: "443"
   container_ports: []
}
```

fancyremarker (approved):
Thanks! Re-tested, all the issues I hit are now addressed.

PR #1092, opened 2025-04-21
Made small QOL improvements for diagnostics dashboards

almathew:
These fixes were identified during a recent customer convo, and all seem legit. 

Including:
* Impacted time range is set in user's local time, and then sent to the server in UTC
* Symptoms are free-form text, not a drop down
* The list of dashboards correctly identifies "Time Created", and adds the "Time Range"

For the final bullet point, this may not be correct anymore when we shift to a single timestamp (timestamp issues were noticed). However, I'm not sure whether we want to display the single timestamp or the range identified by the diagnostic process in those scenarios, so I'm keeping the time range for now.

### DIFF
--- a/src/ui/pages/diagnostics-create.tsx
+++ b/src/ui/pages/diagnostics-create.tsx
@@ -25,6 +25,7 @@ import {
   Button,
   EnvironmentSelect,
   FormGroup,
+  Input,
   Select,
   Tooltip,
 } from "../shared";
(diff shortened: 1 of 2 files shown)


eabruzzese (approved):

fancyremarker:
Re:

> Impacted time range is set in user's local time, and then sent to the server in UTC

I think the request was to be able to configure the time zone via dropdown, not default to local. I don't think defaulting to local is better than defaulting to UTC.

PR #1094, opened 2025-04-22
add site selection for datadog metric drains

michaelwang13:
[sc-32711]

Similar to Log Drains, Datadog Metric Drains can be in different sites. Allow this to be configured on metric drain creation in the UI. 

A difference between this and our CLI `--site` is that this updates series_url regardless. If no site is explicitly stated in the CLI, we assume US1 but don't set series_url.

### DIFF
--- a/src/ui/pages/create-metric-drain.tsx
+++ b/src/ui/pages/create-metric-drain.tsx
@@ -1,6 +1,7 @@
 import {
   type CreateMetricDrainProps,
   type MetricDrainType,
+  datadogSites,
   fetchDatabasesByEnvId,
   provisionMetricDrain,
   selectEnvironmentById,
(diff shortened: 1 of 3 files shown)


joshraker (changes requested):

src/ui/pages/create-metric-drain.tsx
```diff
@@ -40,6 +40,14 @@ const options: SelectOption<MetricDrainType>[] = [
   { value: "datadog", label: "Datadog" },
 ];
 
+const datadogSiteOptions: SelectOption[] = [
+  { label: "US1", value: "US1" },
+  { label: "US3", value: "US3" },
+  { label: "US5", value: "US5" },
+  { label: "EU1", value: "EU1" },
+  { label: "US1-FED", value: "US1-FED" },
+];
```
joshraker, line 49:
```suggestion
const datadogSiteOptions: SelectOption[] = Object.keys(datadogSites).map((site) => (
  { label: site, value: site })
);
```

src/ui/pages/create-metric-drain.tsx
```diff
@@ -105,6 +113,7 @@ export const CreateMetricDrainPage = () => {
   const [dbId, setDbId] = useState("");
   const [handle, setHandle] = useState("");
   const [apiKey, setApiKey] = useState("");
+  const [ddSite, setDdSite] = useState("");
```
joshraker, line 116:
```suggestion
  const [ddSite, setDdSite] = useState("US1");
```

joshraker (approved):

joshraker (approved):

PR #1098, opened 2025-04-25
Add PITR beta page

joshraker:


joshraker:
https://app.shortcut.com/aptible/story/32232/ui-add-pitr-beta-page

src/ui/pages/database-detail-pitr.tsx
eabruzzese, line 1: (Optional/nit) PITR only works for PostgreSQL, right? If so, should we display a message if a user accidentally makes their way to this page for a non-PostgreSQL database? Probably an edge case, but I just did it while I was testing 😅

src/ui/pages/database-detail-pitr.tsx
```diff
@@ -0,0 +1,109 @@
...
+            feedbackMessage="Time in UTC"
+          >
+            <Input
+              className="max-w-xs"
+              name="target-time"
+              value={recoveryTarget}
+              onChange={(e) => setRecoveryTarget(e.currentTarget.value)}
+            />
+          </FormGroup>
+        ) : null}
```
eabruzzese, line 100: (Optional/suggestion) I dunno how robust our timestamp handling is in the backend, but if we're concerned about it at all, I just wanted to point out that we implemented and styled react-datepicker as part of the diagnostics work, which you can see [here](https://github.com/aptible/app-ui/blob/main/src/ui/pages/diagnostics-create.tsx#L189-L200).

Not blocking -- or even suggesting we use it here -- just wanted to point it out in case you hadn't seen that it was available.

src/ui/pages/database-detail-pitr.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const { id = "" } = useParams();
+  useQuery(fetchDatabase({ id }));
+  const dispatch = useDispatch();
+  const navigate = useNavigate();
+
+  const database = useSelector((s) => selectDatabaseById(s, { id }));
+  const [handle, setHandle] = useState("");
+  const [recoveryTargetType, setRecoveryTargetType] = useState("time");
+  const [recoveryTarget, setRecoveryTarget] = useState(
+    DateTime.utc().toFormat("yyyy-MM-dd HH:mm:ss"),
```
eabruzzese, line 32: Should this be ISO8601?

```suggestion
    DateTime.utc().toFormat("yyyy-MM-ddTHH:mm:ssZ"),
```

src/ui/pages/database-detail-pitr.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <div>
+      <Group>
+        <div>
+          Restore this database to a new database and recover the new database
+          to the provided point in time.
+        </div>
+        <div>
+          After clicking "Recover" do not leave this page or the recovery
+          process may not complete.
+        </div>
```
eabruzzese, line 68: (Optional/nit) I wonder if this should be styled as an alert or something?

eabruzzese: In a similar vein: will PITR require the user to update their database URL/credentials? (Sorry I'm a little out of the loop on the backend implementation; just going by the description in the thunk)

If so, should we call that out here somewhere?

src/deploy/database/index.ts
```diff
@@ -722,6 +752,101 @@ export const scaleDatabase = api.post<
...
+      schema.loaders.error({
+        id: ctx.key,
+        message: restoreCtx.json.error.message,
+      }),
+    );
+    return;
+  }
+
+  const restoreData = restoreCtx.json.value;
+  // TODO: POnly wait for the expected database, not the operation to complete.
```
eabruzzese, line 800:
```suggestion
  // TODO: Only wait for the expected database, not the operation to complete.
```

src/deploy/database/index.ts
```diff
@@ -480,14 +493,32 @@ export const createDatabaseOperation = api.post<
...
           disk_size: diskSize,
           provisioned_iops: iops,
           instance_profile: containerProfile,
+          recovery_target_type: recoveryTargetType,
+          recovery_target: recoveryTarget,
+          type,
+        };
+      }
+
+      case "pitr_restore": {
```
eabruzzese, line 515: (Suggestion) Do we also need to add this new operation type to the [filter list on the activity page](https://github.com/aptible/app-ui/blob/main/src/ui/shared/activity.tsx#L263-L282)?

Not sure if there's anything else we need to do there to make sure that those new operations show up.

eabruzzese (approved):
Looks good to me! I know this is a beta so I'm not gonna be real picky. Just a few questions/suggestions, but nothing blocking.

src/ui/pages/database-detail-pitr.tsx
joshraker: > will PITR require the user to update their database URL/credentials?

If they want to switch to the restored database, yes. However, our feedback indicates that most customers would rather reconcile any lost or "bad" data to the existing database instead.

src/ui/pages/database-detail-pitr.tsx
eabruzzese: Yeah. I had copied the wrong database ID (a Redis instance) and put it in the URL.

Totally an edge case. We can address it later if it becomes an issue.

PR #1101, opened 2025-04-30
Require new elasticsearch databases to have at least 1GB of memory

joshraker:
Elasticsearch databases can't start without at least 1GB of memory. This PR removes that option on the create database page when elasticsearch is selected as the type.

### DIFF
--- a/src/ui/pages/create-database.tsx
+++ b/src/ui/pages/create-database.tsx
@@ -20,7 +20,7 @@ import {
   existValidator,
   handleValidator,
 } from "@app/validator";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useNavigate, useSearchParams } from "react-router-dom";
 import { useDatabaseScaler, useValidator } from "../hooks";
 import { EnvironmentDetailLayout } from "../layouts";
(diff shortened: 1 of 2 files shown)


joshraker:
https://app.shortcut.com/aptible/story/32544/when-you-create-an-elasticsearch-database-in-the-ui-it-defaults-to-512-mb-of-ram

src/ui/pages/create-database.tsx
```diff
@@ -108,6 +108,17 @@ export const CreateDatabasePage = () => {
...
+  useEffect(() => {
+    if (imgSelected?.type === "elasticsearch") {
+      if (scaler.containerSize < 1024) {
+        dispatchScaler({
+          type: "containerSize",
+          payload: 1024,
+        });
+      }
+    }
+  }, [imgSelected?.type, scaler.containerSize]);
```
eabruzzese, line 120: (Optional/suggestion) Should we stick a comment here so we don't forget why we're doing this? Maybe declare the `minSize` variable here too (instead of inline in the prop) so all of the elasticsearch-specific stuff is in one place?

eabruzzese (approved):
Looks good! One minor (optional) suggestion

eabruzzese (approved):

PR #1105, opened 2025-05-08
Allow configuration of stop timeout in service deployment settings

aguilinger:
Needs to wait for https://github.com/aptible/deploy-api/pull/1802

Allows user configuration of the stop timeout added in the previous story. I haven't updated the docs, but the docs page already linked will have the longer explanation with caveats of the setting. It's also a little different that the existing checkbox style settings for worker services since it can be set even without an endpoint. I tried to update the language of the popup banner to be clearer there as well.

I also added a save banner to make it more consistent.

With endpoint
Screenshot 2025-05-08 at 7 55 09 AM

Without endpoint
Screenshot 2025-05-08 at 7 55 02 AM


aguilinger:
Added min/max in the text
Screenshot 2025-05-09 at 11 32 43 AM

joshraker (approved):

PR #1107, opened 2025-05-18
Add TZ toggle to the dashboard view page [sc-33100]

almathew:


fancyremarker:
Thanks! Tested and confirms this fixes the following issue from [sc-33069](https://app.shortcut.com/aptible/story/33069):

> If you choose local in the diagnostics generation, you see UTC in the dashboards. You should see local.

One gap is that if you choose local on the dashboard creation page, the dashboard show page will still default to UTC and you'll need to choose local again from a second dropdown. I think that's fine for now.

---

There's another bug (really, pair of bugs with what is likely a single fix) from sc-33069 which this doesn't address. I'll spend a few minutes trying to fix now within this PR (after waiting to ensure I can safely run Hotshot), and if I can't, I'll split into a new ticket:

> - If you choose UTC on dashboard creation, the range in the selector is broken. It only includes the latest local time, meaning the current UTC time isn't in the list for users whose local time is behind UTC
> - Regardless of whether you choose local or UTC on dashboard creation, the time range gets sent (and persisted to `Dashboard#range` as though I'd chosen local). For example, if my local time is UTC-4, I choose "UTC" and a time range of 1:30-2:30pm (13:30-14:30 UTC), the resulting Dashboard will have a range of 17:30-18:30 UTC.

fancyremarker (approved):

fancyremarker:
Added a commit to remove the option for setting Dashboard time range in UTC for now, since it has some bugs, and created [sc-33524](https://app.shortcut.com/aptible/story/33524) to document those bugs for later.

PR #1108, opened 2025-05-19
Remove state validation for Canada on billing method form

joshuaj194:
[sc-32264]

We previously required that a state be selected even when Canada was the chosen country in the billing form. [Confirmed](https://aptible.slack.com/archives/C022BUSF0B1/p1746495292254529?thread_ts=1746473082.801859&cid=C022BUSF0B1) with Finance that we do not require state/province information from Canadian customers.

### DIFF
--- a/src/ui/pages/billing-method.tsx
+++ b/src/ui/pages/billing-method.tsx
@@ -68,7 +68,7 @@ interface FormProps {
   country: string;
 }
 
-const validators = {
+export const validators = {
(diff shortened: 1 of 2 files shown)


benjodo (approved):
- The validator for state fields in billing-method.tsx has been modified to only require state input for US addresses, not for Canadian addresses anymore.
- Added unit tests to verify the validator behavior for different countries, making validators an exported const to enable testing.

PR #1110, opened 2025-06-02
Errors in console

neurosnap:
refactor(stacks): precompute stats by stack and used that for sorting

Instead of calculating a stack stats (e.g. env count, app count, db
count) individually, we can precompute all stacks with their stats
inside of a hash map which will make accessing that data easier and
faster.

### DIFF
--- a/src/ui/pages/stacks.tsx
+++ b/src/ui/pages/stacks.tsx
@@ -14,7 +14,7 @@ import {
   type DeployStackRow,
   selectStacksForTableSearch,
 } from "@app/stack-table";
-import { useMemo, useState } from "react";
+import { useState } from "react";
 import { useSearchParams } from "react-router-dom";
 import { usePaginate } from "../hooks";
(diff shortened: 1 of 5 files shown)


src/ui/pages/stacks.tsx
```diff
@@ -14,7 +14,7 @@ import {
   type DeployStackRow,
   selectStacksForTableSearch,
 } from "@app/stack-table";
-import { useMemo, useState } from "react";
+import { useState } from "react";
 import { useSearchParams } from "react-router-dom";
 import { usePaginate } from "../hooks";
 import { AppSidebarLayout } from "../layouts";
```
benjodo, line 20: Should lines 74-76 refer to `stack` now?

```
<Td variant="center">{stack.envCount}</Td>
<Td variant="center">{stack.appCount}</Td>
<Td variant="center">{stack.dbCount}</Td>
```

PR #1111, opened 2025-06-02
fix(endpoints): replace hostname in credential without a db selector

neurosnap:
Previously our regex was looking for a trailing `/` which only works for connection URLs with a database selector, like postgres.

Many connection URLs do not have a database selector which meant we were not properly replacing their hostnames.

In this revised approach, I use a negative lookahead regex to properly find the last occurence of a `:[0-9]`


benjodo:
Functionally tested locally against an InfluxDB instance ✅

benjodo (approved):

michaelwang13 (approved):
lgtm

PR #1113, opened 2025-06-06
COnfigurable restart free scaling in HAS

aguilinger:
In theory this could be a checkbox but that didn't look right with the rest of the configuration options. The enabled/disabled dropdown also matches i.e. enabled/disabling backups on databases.


neurosnap (approved):
lgtm!

